PR #19, opened 2018-10-10
Documents for copying the provider so it works

symgryph:
Just a quick document to help people setup the provider if terraform doesn't find it.


symgryph:
Hoping this markdown document helps others.

PR #30, opened 2018-12-24
Adds multiple cmd targets for more convenient installation with go get

prologic:
Rather than mucking around with copying identical binaries around ot symlinking them
just provides both and let the `go get` / `go install` tooling do its job :)

Probably should update the README too; but putting this PR up for now to
get feedback.


ggongaware:
I like the idea. If making 2 different binaries, you could optimize the variations to be just ProviderFunc or just ProvisionerFunc in the main.go

prologic:
I can probably do that too -- Gimme a few mins :)

James Mills / prologic

E: prologic@shortcircuit.net.au
W: prologic.shortcircuit.net.au

PR #43, opened 2019-04-12
added config source, edited .gitignore

cardinalit:
Hi!

I tried to use your plugin for Terraform, but I found the missing parameter for the **Proxmox API**
_(Virtual Environment 5.2-1)_. Below, you can see on the screenshot what was the error: 

Screenshot 2019-04-12 at 18 17 47

I fixed it and added a parameter for the API. Have a nice day


ggongaware:
Thanks for the fix!

Issue #55, opened 2019-05-23
"preprovision" must be removed


bkmeneguello:
This "preprovision" stuff is too tied to someone needs. This should be replaced by a "provision" plugin.
The absolute minimum should be make this option "false" by default.

ggongaware:
Preprovision is already skipped in favor of cloud-init. Unfortunately to support older Proxmox versions without cloud-init features, these method are needed to establish connectivity to the other cool built-in provisioners (remote, chef, etc....)

bkmeneguello:
On older versions without cloud-init features it's possible to create ans attach a cloud-init ISO or VFAT handcrafted. It only requires meta-data and user-data files and NoCloud provider will use them (https://cloudinit.readthedocs.io/en/latest/topics/datasources/nocloud.html)

PR #56, opened 2019-05-23
Defaults preprovision to false

bkmeneguello:
#55


ggongaware:
Already skipped when cloud-init... I don't see a real alternative if not using cloud-init being addressed

Issue #32, opened 2018-12-24
Fix deprecation warnings and README samples


prologic:
Managed to fix the deprecation warnings I was seeing with:

```
  disk {
    id = 0
    type = "scsi"
    storage = "zfs"
    size = 8
  }

  cores   = 1
  sockets = 1
  memory  = 2048

  network {
    id = 0
    model = "virtio"
    bridge = "vmbr0"
  }
```

The `README` should be updated to reflect latest changes in mazter.

hbokh:
I had to go a bit further than that...

```txt
  clone   = "jessie-template"
  cores   = 1
  sockets = 1
  memory  = 2048

  network {
    id     = 0
    model  = "virtio"
    bridge = "vmbr0"
  }

  disk {
    id           = 0
    type         = "virtio"
    storage      = "local-lvm"
    storage_type = "lvm"
    size         = "4G"
    backup       = false
  }
```

Also this line is wrong and will result in "400 Parameter verification failed":

`ipconfig0 = "ip=10.0.2.99, gw=10.0.2.2"`

that is missing a CIDR like `<IPv4Format/CIDR>`

so should read something like:

`ipconfig0 = "gw=10.0.2.2,ip=10.0.2.99/24"`

PR #61, opened 2019-06-26
Fix example in README

banzo:
`ipconfig0` parameter does not support a space after the comma and needs the CIDR suffix


banzo:
This "fixes" #32

PR #70, opened 2019-08-05
Add SCSI controller support

V-Paranoiaque:
The new var **scsihw** allows to set an other SCSI Controller model using the Proxmox API


sticky-note:
I know this PR is already merged but there is no doc on README for these two new parameters
@V-Paranoiaque

PR #71, opened 2019-08-05
Feature to set the boot order and the main disk

V-Paranoiaque:


sticky-note:
I know this PR is already merged but there is no doc on README for these two new parameters
@V-Paranoiaque

V-Paranoiaque:
Added :)
https://github.com/Telmate/terraform-provider-proxmox/pull/72
I think the documentation should be somewhere else, with all the parameters and the combinations it will become messy

sticky-note:
The doc is already there -> https://pve.proxmox.com/pve-docs/api-viewer/
I agree with you, this provider needs a radical refresh

PR #75, opened 2019-08-14
support PM_OTP_PROMPT for ondemand interactive OTP

ggongaware:
@V-Paranoiaque can you check this version out?

### DIFF
--- a/proxmox/provider.go
+++ b/proxmox/provider.go
@@ -3,6 +3,7 @@ package proxmox
 import (
 	"crypto/tls"
 	"fmt"
+	"os"
 	"regexp"
 	"strconv"
 	"sync"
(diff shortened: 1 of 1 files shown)


proxmox/provider.go
```diff
@@ -22,6 +23,20 @@ type providerConfiguration struct {
 
 // Provider - Terrafrom properties for proxmox
 func Provider() *schema.Provider {
+	pmOTPprompt := schema.Schema{
+		Type:        schema.TypeString,
+		Optional:    true,
+		DefaultFunc: schema.EnvDefaultFunc("PM_OTP", ""),
+		Description: "OTP 2FA code (if required)",
```
V-Paranoiaque: This description will never be displayed, maybe it should explain how to enable the OTP 2FA support

PR #121, opened 2019-12-04
Do not create resource when clone fail

claudusd:
This PR required the PR 63 on [proxmox-api-go](https://github.com/Telmate/proxmox-api-go/pull/63).

On a previous PR #108 i set the id when I clone VM. 

But when  cloning fail on Proxmox the resource is not created.

But the method cloning method on the promox-api-go clone and configure the vm in the same method.

So i split the cloning and the configuration to set the id only if clone success.


claudusd:
Hi, 

If you didn't merge the proxmox-api-go's PR 63 this wiil not fix the error when configure fail after a clone.

PR #122, opened 2019-12-05
Feature/vcpu

V-Paranoiaque:
Add support for VCPUS in the provider
https://github.com/Telmate/proxmox-api-go/pull/64


ggongaware:
Need to adjust this to default to  empty if less than 1 as well

V-Paranoiaque:
The API is considering a value < 1 as empty, is it not enough @ggongaware?

ggongaware:
That's fine

PR #125, opened 2019-12-08
Add parameters to the disk to set read/write limits

V-Paranoiaque:


cospeedster:
@ggongaware

ggongaware:
Do these need to propagate through the proxmox-api-go?

V-Paranoiaque:
No need @ggongaware the parameters are added automatically by the API

ggongaware:
Do they populate as zeros '0' from the default and is that ok?

V-Paranoiaque:
If the parameter is omitted the parameter is removed by the Proxmox API

PR #148, opened 2020-03-02
WIP: Create additional disks when template cloning

claudusd:
Hi, 

VM creation  doesn't create additional disk when template cloning. And creation fail because the provider try ti resize unknown disk.

This PR add the disk creation for template cloning.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,3 +6,4 @@ require (
 	github.com/Telmate/proxmox-api-go v0.0.0-20191217000250-7338ae30b9b0
 	github.com/hashicorp/terraform v0.12.10
 )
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -6,3 +6,4 @@ require (
 	github.com/Telmate/proxmox-api-go v0.0.0-20191217000250-7338ae30b9b0
 	github.com/hashicorp/terraform v0.12.10
 )
+replace github.com/Telmate/proxmox-api-go => github.com/claudusd/proxmox-api-go cloned_update_config
```
Nothing4You, line 9: I don't think this should be in this PR

PR #152, opened 2020-03-15
Fix macaddr, optional + computed is problematic

frederikbosch:
We have had several fixes already for this problem. The key issue is that `macaddr` is calculated by Proxmox automatically in the majority of cases, but some might want to pass a specific mac address. When using a computed + optional property you have a remaining limitation. 

Suppose a resource was created without supplying the mac address manually. The mac address is thus computed. Then an update is executed. Now the mac address is missing. Because we are using an optional + computed property, Terraform sees this as a missing value, and wants to update the resource. However, there is no update, because macaddr was never set, and is still not set.

What is the solution? To include the intent in the code. Use `mac_manual` if you want a manual address, or have it automatically calculated.

### DIFF
--- a/docs/resource_vm_qemu.md
+++ b/docs/resource_vm_qemu.md
@@ -109,7 +109,8 @@ The following arguments are supported in the resource block:
 * `network` - (Optional)
     * `id` (Required)
     * `model` (Required)
-    * `macaddr` (Optional)
+    * `mac_manual` (Optional) Use if you to pass a manual mac address.
+    * `macaddr` (Computed) Will be either the automatic assigned address by Proxmox or the manual value of `mac_manual`.
     * `bridge` (Optional; defaults to nat)
     * `tag` (Optional; defaults to -1)
     * `firewall` (Optional; defaults to false)
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -175,7 +175,7 @@ func resourceVmQemu() *schema.Resource {
 						"mac_manual": &schema.Schema{
 						    Type:     schema.TypeString,
 						    Optional: true,
-						}
+						},
 						"macaddr": &schema.Schema{
 							Type:     schema.TypeString,
 							Computed: true,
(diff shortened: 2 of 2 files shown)


frederikbosch:
@alterway-ansible Could you have a look at this PR? We need to fix this once and for all.

frederikbosch:
I was thinking, maybe the variable should be `macmanual`, without underscore. We do not seem to have a real guide how to name a variable in the project.

ggongaware:
I like this approach so far. Does it work for anyone who can test it?

winggundamth:
This does not work. I got this error

```
proxmox_vm_qemu.vm: Creating...

Error: Error creating VM: 400 Bad Request, error status: {"data":null,"errors":{"net0":"invalid format - format error\nnet0.mac_manual: property is not defined in schema and the schema does not allow additional properties\n"}} (params: map[agent:1 bios:seabios boot:cdn cores:4 cpu:kvm64 description: hotplug:network,disk,usb ide2:none,media=cdrom memory:8192 name:worker-1 net0:virtio=7E:48:FF:1B:02:60,bridge=vmbr0,mac_manual=DE:73:91:B4:55:21 numa:false onboot:true ostype:l26 scsihw:virtio-scsi-pci sockets:1 virtio0:media=disk,size=100,file=pve-data:88077017/vm-88077017-disk-0.raw,format=raw vmid:88077017])

  on main.tf line 24, in resource "proxmox_vm_qemu" "vm":
  24: resource "proxmox_vm_qemu" "vm" {
```

This is my .tf file

```tf

  name = "worker-1"

  # Node name has to be the same name as within the cluster
  # this might not include the FQDN
  target_node = "pve2"

  vmid = 88077017
  onboot = true
  preprovision = false
  cores = "4"
  memory = "8192"
  scsihw = "virtio-scsi-pci"
  cpu = "kvm64"
  agent = "1"

  # Setup the disk. The id has to be unique
  disk {
    id = 0
    size = 100
    type = "virtio"
    storage = "pve-data"
  }

  network {
    id = 0
    model = "virtio"
    bridge = "vmbr0"
    mac_manual = "DE:73:91:B4:55:21"
  }

  iso = "none"
}
```

PS. I'm using your vmid PR combine with this PR.

frederikbosch:
The PR is updated. At the moment, I cannot test this out. Probably next week. Maybe @winggundamth can try it.

liquid-metal:
Is there any ongoing work on this? I could really use this feature, as I am currently ignoring changes on the entire network block in order to stop terraform from rummaging with my mac addresses
This is less than ideal because it stops me from changing network settings of existing vms with terraform

PR #156, opened 2020-03-17
add vmid to vm qemu resource

frederikbosch:
Completes the feature `vmid` that has been added to LXC.


frederikbosch:
Question: why is there a duplicate name check? Isn't it up to the API to respond duplicate names?

zinefer:
I just tried this out and it works well for me. Thank you!

PR #169, opened 2020-04-30
get ip address of resource_vm_qemu if cloud-init uses dhcp

lausser:
This patch allows to have a config like this:

```
resource "proxmox_vm_qemu" "cloudinit-vm" {
  name = "test.xyz.com"
  clone = "t-centos-7.7"
  agent = 1

  os_type = "cloud-init"
  ipconfig0 = "ip=dhcp"
...
  connection {                  
    user = "packer"                           
    password = var.ssh_password
    // self.ssh_host will automatically be set to the dhcp address
    host = self.ssh_host
  }                                       
  provisioner "remote-exec" {
    inline = [  
      // here you will see the actual ip address        
      "/sbin/ip a"  
    ]      
  }
```

The code was partly copied from https://github.com/hashicorp/packer/blob/master/builder/proxmox/builder.go where dhcp+qemu-guest-agent are the only way to connect to the vm.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -1046,9 +1046,57 @@ func initConnInfo(
 			sshHost = d.Get("ssh_forward_ip").(string)
 		}
 		if sshHost == "" {
-			// parse IP address out of ipconfig0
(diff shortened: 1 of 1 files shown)


faragm:
I got the following error when with this configuration
`Error: error unmarshalling result Could not parse fe80::61d7:7545:7803:c63b%6 as IP`

PR #184, opened 2020-06-21
Added SSD emulation and Discard options. Updated docs and examples

nzolot:
This PR adds Discard option for disks and includes changes form PR #146  + docs

### DIFF
--- a/docs/resource_vm_qemu.md
+++ b/docs/resource_vm_qemu.md
@@ -128,6 +128,8 @@ The following arguments are supported in the resource block:
     * `backup` (Optional; defaults to false)
     * `iothread` (Optional; defaults to false)
     * `replicate` (Optional; defaults to false)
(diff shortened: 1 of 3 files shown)


ggongaware:
One error to clear up first - proxmox/resource_vm_qemu.go:286:30: Errorf format %d has arg v of wrong type string

nzolot:
> 
> 
> One error to clear up first - proxmox/resource_vm_qemu.go:286:30: Errorf format %d has arg v of wrong type string

Thanks @ggongaware , I've missed travis test results. Looks good now

nzolot:
After some testing in prod environment, I've decided to remove default values for 'ssd' and 'discard' options.

Main reason for this is the case when one have configured VMs using previous version of provider without ssd and discard options. In this case provider will change configuration of existing disks, but this change cannot be applied as 'hotplug'. As a result there will be VMs that have many 'pending' changes and will require stop-start.

In order to use 'ssd' and 'discard' options, now they should be configured explicitly in terraform

Issue #112, opened 2019-11-12
Terraform wants to change mac address on reapply


cospeedster:
# Problem
When I plan the second run of terraform, it want to change the network blocks:
```
...

      + network {
          + bridge    = "vmbr0"
          + firewall  = false
          + id        = 0
          + link_down = false
          + model     = "virtio"
          + queues    = -1
          + rate      = -1
          + tag       = 2101
        }
      - network {
          - bridge    = "vmbr0" -> null
          - firewall  = false -> null
          - id        = 0 -> null
          - link_down = false -> null
          - macaddr   = "42:CE:92:97:95:4B" -> null
          - model     = "virtio" -> null
          - queues    = -1 -> null
          - rate      = -1 -> null
          - tag       = 2101 -> null
        }
      + network {
          + macaddr = "42:CE:92:97:95:4B"
        }
    }

Plan: 0 to add, 2 to change, 0 to destroy.
```

Maybe this Issue has same cause for Issue #47. @ggongaware Mentioned this here https://github.com/Telmate/terraform-provider-proxmox/issues/47#issuecomment-494627802.

# Work around 

For a work around I added this to my resource:
```HCL
  # For some reason terraform has changes on reapply
  lifecycle {
    ignore_changes = [
      network,
    ]
  }
```
I don't know how I can just ignore changes on mac addresses.

### Any Ideas there so far?

cwiggs:
I had this same issue and adding your workaround does "solve" the problem for now, but isn't ideal.

An interesting thing about this bug is that when I tried to `terraform apply` without the `lifecycle` workaround in place terraform fails to apply, with this error: `Error: 400 Parameter verification failed.`

Sounds like it's getting a 400 back from the proxmox API.

cospeedster:
@cwiggs I totaly agree, that this isn't the best way. That's why I opened this issue here.

cwiggs:
@cospeedster I didn't mean to come off negative towards your workaround, sorry if I did.  Thanks for the workaround and this issue!

I enabled terraform debugging and ran terraform with the `lifecycle` workaround commented out.  the debug log is around 2k lines so I don't want to include the whole thing, but here is the response from the proxmox api with the 400 response:

```
2020-01-21T10:05:01.976-0700 [DEBUG] plugin.terraform-provider-proxmox: 2020/01/21 10:05:01 <<<<<<<<<< RESULT:
[...]

Issue #88, opened 2019-09-16
Idempotency problem - Mac address must be left to computed. 


helldorado:
hello, 
first of all, thanks for this great provider. 

A major regression was introduced with this change form commit https://github.com/Telmate/terraform-provider-proxmox/commit/bf98b941f7bdfea2d4c19239d04d4b57c19f9da6 

This completely breaks the provider's idempotency and is unusable in production.

Some issues like #82 is related to this commit.

I don’t see the point of generating or providing a MAC address while Proxmox API already does the job and returns the result in its answer.

I patched locally for our needs. Is there any chance to rollback to `Computed: true` ?
If this is not possible, a refactoring of the function https://github.com/Telmate/proxmox-api-go/blob/master/proxmox/config_qemu.go#L661-L704 will be necessary, and we can make a patch for the community.

Best regards.

mbuiro:
You can sort of get around this by hardcoding the MAC address yourself:

```
    # Setup the network interface 
    network {
        id = 0
        model = "virtio"
        bridge = "vmbr0"
        firewall = false
        macaddr = "DE:1E:8C:42:E8:52"
    }
```

Unfortunately, the next idempotency issue stems from the cloud init password:

```
# proxmox_vm_qemu.cloudinit-test will be updated in-place
  ~ resource "proxmox_vm_qemu" "cloudinit-test" {
        agent        = 1
        boot         = "cdn"
        bootdisk     = "virtio0"
      ~ cipassword   = "**********" -> "supersecretpassword"
```

alexanderek:
> Unfortunately, the next idempotency issue stems from the cloud init password

How can this be fixed?

cospeedster:
I mentioned a work around here: Telmate/terraform-provider-proxmox#112

PR #198, opened 2020-07-26
Various idempotency fixes, update to new terraform sdk, addition of logging library

doransmestad:
Hey all,

A couple months ago I went to use this provider to control my HA proxmox cluster with terraform.  Really awesome job getting this plugin together!   I was super excited to see someone spent the time and effort to build this.  Unfortunately I hit a few issues with idempotency, so I decided to learn how to build terraform providers so I could get this working for myself and contribute back.

Well.. one thing led to another and I found myself changing a number of things all at once to fix my needs.  It is now working well for me, but it appears I have merge conflicts with master branch and so I'm marking this PR as a work in progress.  Before I spend the time to get everything back in sync, I wanted to reach out to check if my changes would be welcomed?  

Changes:
- Changed the definitions for `disk` and `network` to TypeList to eliminate idempotency issues (especially around macaddrs as described in #112 and #88).  Added the appropriate expand and flatten functions to handle pxapi <-> terraform type translations.
- Updated the provider to use the new terraform SDK library (github.com/hashicorp/terraform-plugin-sdk) in preparation for updating to the upcoming v2 SDK.
- I updated a few attributes to be computed instead of having a default value set:  `scsihw` `searchdomain` `nameserver` and `bootdisk`.  This was needed to prevent issues cloning a template that had these set.
- I changed the Update and Create functions to return a call of Read as recommended by the SDK, this appears to have resolved a number of other idempotency issues.
- To aid in debugging, I added a full logging library to the plugin that can be enabled if desired using provider-level options.
- To make it easier to share, I added github actions which automatically build the provider binary on all branches.

Breaking Changes:
- I had to remove the provisioner code as the new terraform SDK no longer supports functions it depended upon (sorry).  This might yet be made to function, but I haven't investigated how to do so.
- I removed a number of depreciated attributes from the resource schema

Again, this got a bit out of hand from my original intent...  but the good news is I'm now able to use the plugin without idempotency issues.  If I were to clean up this Pull Request (and add documentation), would there be interest in adopting the changes here?   Perhaps as a v2 of the plugin due to the breaking changes I listed above?

If not, I totally understand.   Please let me know your thoughts.

Thanks.


ggongaware:
I think your changes are great!

I haven't been able to keep up with version updates trying to maintain clusters of older versions of Proxmox deployments. 

I'd like to make your update the master branch, and move the version I still use into a legacy branch.

Also, another developer started a new CI approach to validating against an actual Proxmox host:

https://github.com/Telmate/proxmox-api-go/pull/84

If we can get your changes + a CI that proves a baseline functionality is working, that will be the new gold standard.

doransmestad:
Awesome!  Okay, give me a couple days to clean the PR up a bit and add docs.  I'll let you know when it's all set.

Also- super cool to hear about the vagrant environment for testing!  That's a clever way to do it.  It would be sweet to get some testing in.  Thanks for linking that, I'll take a closer look and see if I can tackle that after this (if someone else doesn't beat me to it).

Aaron-K-T-Berry:
Super cool stuff, I especially appreciate the logging changes as I just went through the effort of adding my own logs everywhere and this is much better 🎉

doransmestad:
Thanks @Aaron-K-T-Berry!   Glad you find the logging useful!   I went back and forth on whether to build it out, but ultimately it really helped me learn how things worked and then debug later on.  :smile: 

@ggongaware   I pushed up an update to the PR.  It is now rebased off the top of master branch.  I purposely used a "2 commits old" version of the proxmox-api-go library because it looks like #201 adds that timeout functionality in already.

To make the updates clear vs the original commit, I kept these changes in a second commit, but let me know if you'd prefer I squashed down to just 1 commit.   The changes are:

- Removed an incorrect limitation from the README that slipped in my first commit
- Removed docs on the provisioner code as this PR removes that (due to the SDK not supporting it in its current form)
- Added documentation about the new logging to the provider.md file
- Added docs about the new resource attributes to resource_vm_qemu.md file
- Ran `go fmt`  which corrected a whole bunch of spacing.. thought I did that the first time around, but apparently not.  :sweat_smile: 

I tested creation, update, and deletion actions with my proxmox cluster and all looks good on my end.  Please do let me know though if any tests on your end show anything odd!

doransmestad:
Found a small bug:  forgot to update the Makefile to create the provider as "bin/terraform-provider-proxmox_v2.0.0".  If I recall correctly, the version tag at the end is how terraform knows the plugin uses the newer version of the SDK (v1).  I was `cp`ing it around manually and thus forgot to update the Makefile the first time around.   https://www.terraform.io/docs/configuration/providers.html#plugin-names-and-versions.

With that, I think this is good to go so I'll remove the WIP from the title.

mleone87:
Hii @doransmestad  very very nice work!!!
I got your latest pull request code ad tried on my setup.
Creating(cloning) a cloud init template works as usual, but I noticed that editing it(simply renaming the vm), does not complains about the network part but the mac address get changed anyway in the process, so vm network is unreaceable until a forced reboot

Any ideas why?

doransmestad:
Thanks @mleone87  and thank you for running the test!   To make sure I understand, your saying the network mac address changes when you:

1)  Create a new qemu resource
2)  Edit the qemu resource (a simple change, such as the vm name)
3)  Although the terraform diff doesn't indicate a change, the mac address on the network cards are changed in proxmox

If so.. then yep, I was able to reproduce that issue.  Very odd.  I'm not sure why that would be happening  :thinking:   I'll dive in here.  My first suspicion is the update request we're sending via the proxmox-api-go library is changing more than intended, but it could easily be an issue with my changes too.   Thank you for letting me know!   I'll investigate and post an update shortly.

doransmestad:
Well, my guess was totally wrong: this is definitely my bug.  :laughing:   Great find @mleone87:  this stems from an incorrect d.Set(..) call I have which was given more attributes than were available in the resource definition (in particular proxmox-api-go passes back a device Id that is needed, but I never stripped out after using it).   

Fixing that uncovered another issue with idempotency around disk entries (particularly around storage_type) that I'm working through now.  I'll get this PR fixed within the next couple days or so.

mleone87:
thanks @doransmestad, you mean this part of code(around line 1070)?
	// Deprecated single disk config.
	d.Set("storage", config.Storage)
	d.Set("disk_gb", config.DiskSize)
	d.Set("storage_type", config.StorageType)
	// Deprecated single nic config.
	d.Set("nic", config.QemuNicModel)
	d.Set("bridge", config.QemuBrige)
	d.Set("vlan", config.QemuVlanTag)
	d.Set("mac", config.QemuMacAddr)
	d.Set("pool", vmr.Pool())

doransmestad:
Close!   It's a right above that:

The buggy code (currently line 1068 in this PR), does this:

```
flatNetworks, _ := FlattenDevicesList(config.QemuNetworks)
d.Set("network", flatNetworks)
```

The issue here is that d.Set(..) can't set `flatNetworks` into the resource because there is an attribute `id` that the network schema.TypeList doesn't have defined (in the resource definition around line 275).  I commented `id` out as that is now a built-in aspect of the flatNetworks list and I figured asking the user to set it was redundant.  In other words `id` = 0  instead becomes  flatNetworks[0]

That works great... except I didn't remove the `id` attribute from the map, so terraform threw an error.. which I also did not catch  :facepalm: 

Fixed the code looks like this:

```
// flatten the structure into the format terraform needs and remove the "id" attribute as that will be encoded into
// the list structure.                                                                                             
flatNetworks, _ := FlattenDevicesList(config.QemuNetworks)                                                         
flatNetworks, _ = DropElementsFromMap([]string{"id"}, flatNetworks)                                                
if err = d.Set("network", flatNetworks); err != nil {                                                              
        return err                                                                                                 
}                                                                                                                  
```

Now it works correctly AND it'll also properly catch any future errors and correctly inform terraform to stop  :smile:    The disk idempotency thing though.. I'm still not sure on.   One issue I'm seeing is that I'm apparently not getting storage_type back from proxmox-api-go  (or at least something in my code is squashing it).   I get a response back from `pxapi.NewConfigQemuFromApi` that looks like this:

```
(just showing disk entry here:)

"disk":{"0":{"file":"vm-123-disk-0","format":"raw","id":0,"media":"disk","size":"10444M","storage":"ceph-ssd","type":"scsi"}}
```

but I had expected `storage_type` to also be in that map (as I specified it during creation).. and it isn't there.  So I'm missing something, and terraform thinks the whole disk needs to be replaced.. thus the investigation continues.  :)

mleone87:
@doransmestad Thanks, I hadn't thought about that line of code because I imagined a whole different flow, in fact I still don't explain to myself why those lines of code are present if we set everything through 'network'.
I guess the **DropElementsFromMap** function is your secret since it's not included in your pr code? :smile:
Anyway, what are you referring to with the disk resizing problem? With the code of your last pr I can get a disk resize without freeze of the machine, but I lose connectivity since the mac address is always and in any case changed for the problem you indicate.
Thats the api calls that I got from a terraform apply after i change size in the vm definition(note the POST config that changes mac address and the PUT resize that resizes the vm disk)
qemu_resize

doransmestad:
> I guess the DropElementsFromMap function is your secret since it's not included in your pr code? 

@mleone87   Hah!  I didn't mean to keep that secret.. I totally meant to push my latest changes shortly after posting, but other priorities took over my schedule for a bit there.   Just pushed the new commit with (hopefully) all these issues resolved.   The new commit has:

1)  The fix for the incorrect network updates as I described in my earlier comment

2)  The fix for 'disk' sections not getting updated at all (oops).  This is related to the network update issue as I was also not removing the "id" attribute from 'disk' either which caused problems.

3)  The fix for the disk idempotency issue I alluded to earlier:   with 'disk' updates corrected, I ran into an idempotence issue with `storage_type` as it is required by proxmox-api-go for creation/updates, but it is not returned in reads (as it appears to only be used for correctly format the disk "file" attribute).   So terraform would remove `storage_type` from the state (as it wasn't returned by proxmox-api-go), but then terraform would want to set it again during the next `terraform apply` since it is in the `.tf` file provided by the user.   To work around this, I used the pre-existing state to carry the `storage_type` over  (see https://github.com/Telmate/terraform-provider-proxmox/pull/198/commits/24d839bc925b4f8f43f44a2f8326edf5ab178bda#diff-0b068297509ee768930209ff57f2bddbR1068).

That worked nicely, but then caused a crash in proxmox-api-go when attempting an update.  After a long debugging session I discovered that sending "file" and "media" attributes for disks to proxmox-api-go during an update caused the library issues.  I removed those attributes when updating (see https://github.com/Telmate/terraform-provider-proxmox/pull/198/commits/24d839bc925b4f8f43f44a2f8326edf5ab178bda#diff-0b068297509ee768930209ff57f2bddbR866)  and now things are looking good!

Apologies for the long explanation here, this probably all makes sense to you already!  The long winded stuff here is mostly for my own benefit to act as notes for my future self (and perhaps helpful to some future contributor).   :smile:

doransmestad:
Hmm.. great catch @mholasek.  This looks to be a bug of sorts, but I must admit I'm not sure why.  I'm running Ceph on my end so the error I receive is different;  terraform says it successfully completed, but I get a task error in proxmox, and the change wasn't actually applied:  

> TASK ERROR: rbd error: rbd: error opening image vm-131-disk-1: (2) No such file or directory

So I'm guessing there is some necessary step I'm missing to create/init a new disk on the proxmox end prior to updating the VM configuration.  I'll look into this!

Also.. I totally should have tested two disks before pushing that last commit.. I'm a bit embarrassed I didn't think to do that!  Oops.  Thank you for doing so!    In addition to the multiple disk issue you found, I think I'm going to also take a look at adding in integration tests using the terraform SDK so we can begin to build a set of repeatable tests to catch cases like this as part of the development process.  I'll hopefully be able to push a new commit in a few days.

yukron:
@doransmestad, did you have a chance to update this? I was struggling with the current master branch agains latest API and latest Terraform, got some dirty fixes here and there, but your work is way greater and introduced lot of changes. Hope to see it merged someday. :)

doransmestad:
Hey @yukron  thanks for reaching out!  I have made some fixes locally, but unfortunately I had to step away for a bit and deal with pressing issues that took up my time.  That should be wrapping up this week though and I'll be able to get back to this next week!  

Looks like you grabbed the changes here and made another PR with a sync to latest master; awesome stuff!   Do you want to take over and I close this PR out and we use the one you made?   Or, I'm happy to incorporate your changes into this branch and then continue forward?   I don't mind either way!   We'll all benefit no matter the approach.  :smile:

yukron:
Thanks @doransmestad , I think let's wait for your new fixes anyway :) I can maintain my PR until you come back, since there are just your fixes+updates to upstream master. BTW, there is also a different branch based on these changes, which adds some Read function checks to populate undefined vars coming from Proxmox API with 'null' values: https://github.com/yukron/terraform-provider-proxmox/tree/feature/populate_keys_on_import_with_default

I'm also looking into adding additional API calls in proxmox-api-go to populate 'storage_type' and 'format' directly from Proxmox:
https://github.com/yukron/proxmox-api-go/tree/feature/added_getstoragestatus_api_request

doransmestad:
Hey @yukron!   I'm back!   First, a HUGE thank you for the rebase on top of the latest master branch!  I'm going to restart my branch using yours as the source; it all looks great.

I'll also take a look at your branch for populating default values-  I would have thought terraform would do that automatically for us, but if you needed to add them in explicitly I'll take a closer look at the build-in behavior and adopt those in.

I'm super excited about your change to the proxmox-api-go project as well!  Having storage_type provided will allow me to remove my earlier hack around it.  And I see it was just merged yesterday.  Very cool.

-

So my plan here:
1.   I'm catching back up and also adopting the new SDKv2 from terraform.  There are a couple cleanups but it's looks like a nice and smooth upgrade from the v1  :tada:  :heavy_check_mark: 
2.   Fix any immediately apparent idempotency issues (namely, the disk size - more on that shortly) so we have something working in the short term, even if other bugs might be present   :heavy_check_mark: 
3.   Add a true integration test framework in using the SDK provided functions.  This will probably be a bit of work, but once done it should greatly accelerate future work
4.   Squash as many bugs I can within a "reasonable" time adding tests along the way.  I don't want to take too much time making this perfect as we'll be here forever, but squash enough bugs so we're comfortable with a merge.
5.   Finally get this merged in.   

Known bugs so far:
* specifying multiple disks (at least after a clone) causes chaos
* destroying a VM which is managed by HA has no effect (terraform completes, but the VM isn't destroyed)

Fixed:
* idempotency issue around disk size.  long story short:  the proxmox-api-go library only exposed a method for increasing disk size by 1GB increments.  This caused issues because it was impossible to "set" the disk size as one would expect from terraform. To fix this I proposed a new function to proxmox-api-go allowing the caller to set the disk size directly thus eliminating the 'calculate the right size to increase by' quagmire here.  https://github.com/Telmate/proxmox-api-go/pull/95

doransmestad:
> Add a true integration test framework in using the SDK provided functions. This will probably be a bit of work, but once done it should greatly accelerate future work

Done!  :tada:    Well, okay, more like I finished just the start of a framework as only have a single test case in here so far.  It took way more time than it should have because the terraform SDK documentation leaves much to the imagination.  Had to pull inspiration from the terraform-aws-provider to finally get the syntax correct.   On the positive side though, this single test case did immediately find a bug caused by a debug line I added in awhile back!   One bug down, many more to go I'm sure.  :smile: 

Next up:   I plan to add sufficient test cases to cover the most critical code paths, and add in helper functions to test updates correctly applied in proxmox itself (vs the current test setup which is only able to compare with the terraform state itself).

mholasek:
@doransmestad : I appreciate that you are working on improving and expanding the PVE terraform provider. If you need help with testing, please let me know. I am happy to help. We are already waiting for the final version to be ready for use.

zimmertr:
Thanks everyone for the work on this! Very much appreciated!

doransmestad:
Thanks for the merge @ggongaware!   Exciting!   @mholasek  really appreciate the offer for the testing help;  now that this is merged to master, please feel free to open new issues in the repository with anything you find-  I'll be around and will submit fixes when I can!

PR #225, opened 2020-10-19
incorporated doransmestad changes with latest upstream fixes

yukron:
This is just an updated PR #198  of @doransmestad with latest upstream changes.


Matchlighter:
It seems like most plugins that use the new SDK use `func Provider() *terraform.ResourceProvider {` instead of `func Provider() *schema.Provider {`.

proxmox/resource_vm_qemu.go
```diff
@@ -223,15 +321,11 @@ func resourceVmQemu() *schema.Resource {
 				},
 			},
 			"disk": &schema.Schema{
-				Type:          schema.TypeSet,
+				Type:          schema.TypeList,
 				Optional:      true,
 				ConflictsWith: []string{"disk_gb", "storage", "storage_type"},
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
-						"id": &schema.Schema{
```
Matchlighter, line 231: From my understanding of the code and PVE API, I think removing this will have unintended side-effects. For example, what if, in my Terraform file, I have 3 disks - 1, 2, & 3. With this change, if I remove disk 2, it ends up mount the old disk 2 volume in the disk 3 (now 2) slot.

Matchlighter:
I've got a change that I'm working on as well. It'll definitely be a breaking change compared to origin, but you may be interested (it's based on your fork) - I'll upload a WIP branch to my fork soon. I'm mainly making things more consistent, but I also plan on adding a separate resource for disks, kind of like the `aws_instance` has an option to nest EBSs, but also via a separate resource.

Matchlighter:
For reference:
https://github.com/Matchlighter/terraform-provider-proxmox/commits/master
https://github.com/Matchlighter/proxmox-api-go/commits/master

PR #260, opened 2020-11-22
Fix relative links to providers

jdmarble:
These were 404ing from the GitHub view. Looks like they need to be relative to this file's location.


ggongaware:
Changing the links will break the terraform registry view> https://registry.terraform.io/providers/Telmate/proxmox/latest/docs

Any way to have it work for both?

jdmarble:
Oh! I didn't think about that. I can't think of a way to resolve it for both and it's easy enough to manually navigate to the provider docs in the GitHub interface. Just going to forget about it.

Issue #272, opened 2020-12-14
Set network info with ipconfig0


AcidCrash1337:
I've been working on this for about 6 hours now, repeatedly rebuilding my base image (ubuntu20.04) with different options to try to figure out what I need to do to get ipconfig0 to set my IP address.

The key line in my preseed file is this:
```
d-i pkgsel/include string open-vm-tools openssh-server curl python3-pip cloud-init
```

My main.tf file looks very similar to [the simple cloudinit example](https://github.com/Telmate/terraform-provider-proxmox/blob/master/examples/cloudinit_example.tf)
```
variable "SSH_KEY" {
  default = "ssh-ed25519 AAAAC3SNIP"
}
variable "ip_address" {
  default = "SNIP.50"
}
variable "cidr" {
  default = "24"
}
variable "gateway" {
  default = "SNIP.1"
}
variable "nameservers" {
  default = "SNIP SNIP"
}

terraform {
  required_providers {
    proxmox = {
      source = "Telmate/proxmox"
    }
  }
}

provider "proxmox" {
}

resource "proxmox_vm_qemu" "proxmox_vm" {
  count             = 1
  name              = "ns1.DOMAIN"
  target_node       = "SNIP"
  clone             = "ubuntu-20.04"
  full_clone        = false  # Linked clone
  os_type           = "cloud-init"
  cores             = 1
  sockets           = "1"
  cpu               = "host"
  memory            = 512
  scsihw            = "virtio-scsi-pci"
  bootdisk          = "scsi0"
  disk {
    size            = "20G"
    type            = "scsi"
    storage         = "zfsdisk"
    iothread        = true
  }
  network {
    model           = "virtio"
    bridge          = "vmbr0"
  }
  # Cloud Init Settings
  # Reference: https://pve.proxmox.com/wiki/Cloud-Init_Support
  ipconfig0 = "ip=${var.ip_address}/${var.cidr},gw=${var.gateway}"
  nameserver = var.nameservers
  sshkeys = <<EOF
var.SSH_KEY
SNIP
  EOF
  # Connection info for all of the provisioners to use
  connection {
    type = "ssh"
    host = var.ip_address
    #host = "SNIP.207"
    user = "root"
    private_key = file(pathexpand("~/.ssh/id_ed25519"))
  }

  # Provisioners
  provisioner "file" {
    source = "named.conf.options"
    destination = "/etc/bind"
  }

  # More provisioners here
```

[...]

PR #275, opened 2020-12-16
Improved Cloud-Init documentation

AcidCrash1337:
Documentation about Cloud-Init so nobody ever has the same experience I had in https://github.com/Telmate/terraform-provider-proxmox/issues/272


ggongaware:
Thank you for the detailed documentation!

Issue #283, opened 2021-01-03
proxmox insists on creating a new disk when update cloud-init/qemu resource


mgebundy:
Despite there only being a single disk defined in hcl, when I update the resource, a brand new disk is created, breaking the vm.

Here's the resource definition I start with:

```hcl
resource "proxmox_vm_qemu" "vm" {
    depends_on        = [
        null_resource.vm_user_data
    ]
    name              = "vm1"
    target_node       = "home"

    clone             = "vm-cloud-init-template"

    os_type           = "cloud-init"
    cores             = 1
    sockets           = "1"

    cpu               = "host"
    memory            = 4096
    scsihw            = "virtio-scsi-pci"
    bootdisk          = "scsi0"

    cicustom = "user=local:snippets/user-data.yml"

    disk {
        size            = "5G"
        type            = "scsi"
        storage         = "local-lvm"
        iothread        = true
    }

    network {
        model           = "virtio"
        bridge          = "vmbr0"
    }

    lifecycle {
        ignore_changes  = [
            network,
            desc
        ]
    }
}
```

When I make a change to the resource, such as ONLY bumping `cores` from 1 to 2, a new disk is created.

![image](https://user-images.githubusercontent.com/584925/103487305-520d3400-4dd2-11eb-9a33-f39b5c928963.png)

Adding "disk" to the `ignore_changes` list doesn't seem to make any difference, I'm stumped.

mleone87:
can you make some test with a simpler config? this issue should be fixed

mgebundy:
> can you make some test with a simpler config? this issue should be fixed

can you please point to where this is fixed? I can try making a simpler config, but to what end?

mleone87:
> 
> 
> > can you make some test with a simpler config? this issue should be fixed
> 
> can you please point to where this is fixed? I can try making a simpler config, but to what end?

PR #266 and #253 should have fixed this issues, the acceptance test in the cose works correclty with cloning/updating disk resources.

You could also delete all terraform files and make it download latest version of api and go provider

Also, what version of proxmox do you use?

mgebundy:
Thanks.

I'm using the latest version of the provider that exists in the terraform registry, 2.6.6, I expect both those fixed made it into that release. I'm using Proxmox VE 6.2.

mgebundy:
I think I've narrowed it down to the `disk.iothread`, if I don't set that value it, the drive isn't recreated.

I can reproduce the issue with this configuration:

[...]

PR #286, opened 2021-01-08
QEMU Resource Docs Update

KingKeithC:
This PR updates the QEMU resource documentation with data types, default values, and references to official Proxmox documentation where useful. This also organizes all the arguments and attributes into tables, rather than using bullet points, as (IMHO) it makes it easier to scroll through and view at-a-glance which options are available.

I have done my best to read through the code to determine how the arguments are processed by the provider, the proxmox-api-go lib, and the PVE API, so the descriptions can be as useful as possible. If I missed anything, or any of the descriptions are inaccurate, please feel free to suggest any all changes amendments you feel are necessary.

### DIFF
--- a/docs/resources/vm_qemu.md
+++ b/docs/resources/vm_qemu.md
@@ -84,14 +84,12 @@ The following arguments are supported in the resource block.
 |--------|----|---------|-------------|-----------|
 |`name`|`string`|**Yes**||The name of the VM.|
 |`target_node`|`string`|**Yes**||The name of the Proxmox Node on which to place the VM.|
-|`desc`|`string`|**Yes**|``||
-
 |`vmid`|`integer`|No|`0`|The ID of the VM in Proxmox. The default value of `0` indicates it should use the next available ID in the sequence.|
 |`desc`|`string`|No|`""`|The description of the VM. Shows as the 'Notes' field in the Proxmox GUI.|
 |`define_connection_info`|`bool`|No|`true`|Define the (SSH) connection parameters for preprovisioners, see config block below.|
 |`bios`|`string`|No|`"seabios"`|The BIOS to use, options are `seabios` or `ovmf` for UEFI.|
 |`onboot`|`bool`|No|`true`|Whether to have the VM startup after the PVE node starts.|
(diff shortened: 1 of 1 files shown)


docs/resources/vm_qemu.md
```diff
@@ -69,108 +61,178 @@ EOF
...
+
+|Argument|Type|Default Value|Description|
+|--------|----|-------------|-----------|
+|`type`|`str`||**Required** The type of disk device to add. Options: `ide`, `sata`, `scsi`, `virtio`.|
+|`storage`|`str`||**Required** The name of the storage pool on which to store the disk.|
+|`size`|`str`||**Required** The size of the created disk, format must match the regex `\d+[GMK]`, where G, M, and K represent Gigabytes, Megabytes, and Kilobytes respectively.|
+|`format`|`str`|`"raw"`|The drive’s backing file’s data format.|
+|`cache`|`str`|`"none"`|The drive’s cache mode. Options: `directsync`, `none`, `unsafe`, `writeback`, `writethrough`|
+|`backup`|`bool`|`false`|Whether the drive should be included when making backups.|
+|`iothread`|`bool`|`false`|Whether to use iothreads for this drive. Only effective with a disk of type `virtio`, or `scsi` when the the emulated controller type (`scsihw` top level block argument) is `virtio-scsi-single`.|
```
mleone87, line 196: this parameter(maybe some others) should be an integer, ad discovered in issue #283 

api

KingKeithC: I agree, on the backend `iothread` (and a couple other parameters,) should be `int`s. However they are currently defined as `bool`s in the resource schema, so if you were to supply an `int` in the terraform configuration, terraform would return a validation error.

https://github.com/Telmate/terraform-provider-proxmox/blob/944cd338944b838c842263e42c57364b97be9c82/proxmox/resource_vm_qemu.go#L385-L389

A separate PR is needed to either change the schema to `int`s, or they could remain as `bool`s (since that's effectively what `1` and `0` represent anyway,) and the provider could be changed to cast the `bool`s to an `int` before passing the parameters to the PVE API.

mleone87: Correct, I've made the PR to change types of
- iothread
- backup
- replicate
- ssd

from bool to int.
I would stick to API types instead of doing hacks in provider, but feel free to modify if I forgot somthing or if my approach is wrong :)

Issue #256, opened 2020-11-19
Vm reboots after any changes in template


scrowbar:
I've just found that  commit 6c27554  forces VM to be rebooted even after minor changes in template.

For example we have main.tf

```
resource "proxmox_vm_qemu" "testvm" {
   ...
   name        = "test"
   ...
}
```

Then we change name = test2 and perform plan and apply

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  ~ update in-place
...
      ~ name                   = test" -> "test2"
...
```
AFAIK Some changes in proxmox are not require reboot VM. 

Is this expected behavior ? Can we make reboot optional ?

msbence:
This is a blocker for me as well :( Most of the settings definitely not require a reboot to apply.

sophiexiaoyan:
This is also a blocker for me. For example, adding more disks or increasing disk size should not require a VM reboot.

msbence:
I might make a PR about it, but if anyone know which changes require a reboot exactly, that would be a huge help.

msbence:
Okay, made the PR, but feel free to take a look at the code :)

scrowbar:
Thank  you!

PR #292, opened 2021-01-15
Reboot only when necessary + better shutdown logic

msbence:
Currrently any changes made to vm_qemu resources triggers a stop&start task for the VM.
This is not always necessary, for example you can rename a VM while it's running.

My changes: I introduced a new variable in the schema called `reboot_required`. Whenever a "critical" configration parameter changes, this variable will be set to `true`. The end of the resource update will power cycle the VM only if this variable is set to `true`.
Regarding shutdown: currently, it stops the VM instead of a graceful shutdown. This always works, but definitelly not safe in production environment. So I rewritten that part to try a graceful shutdown first, and if that fails it will try to stop the VM forcefully.

This should fix issue #256 as well.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -668,6 +668,11 @@ func resourceVmQemu() *schema.Resource {
 				Optional: true,
 				ForceNew: true,
 			},
+			"reboot_required": {
+				Type: schema.TypeBool,
+				Computed: true,
+				Description: "Internal variable, true if any of the modified parameters require a reboot to take effect.",
+			},
 		},
 	}
 	return thisResource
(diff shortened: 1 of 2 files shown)


mleone87:
Thanks for your PR @msbence !
The logic is ok to me but i think that the reboot required parameter should be only internal, there is no point in setting it in the vm parameters if you dont want to force override it

so, i would only set a "force_reboot" parameter that overrides the normal logic that you use and change the rest according

msbence:
Thank you @mleone87 for taking time to review it!
I'm no expert in terraform provider developing, I borrowed the idea from [Hashicorp's VMware provider](https://github.com/hashicorp/terraform-provider-vsphere/blob/master/vsphere/resource_vsphere_virtual_machine.go#L257).

If I understand correctly, you can not override the parameter at all because it's computed.
If you want so, I can modify my code, but I've failed to find the difference between the two approach.

mleone87:
sorry @msbence  i completely misread the parameter and failed to understand the rest! It makes a lot of sense now!!

Do you have the chance to test this code with some acceptance tests?
You should be able to make some modifications to **resource_vm_qemu_test.go** to match your environment(**testAccProxmoxTargetNode** and the functions returning a sample conf)

msbence:
Yes, I should write the tests as well. However, I have to look into how testing works, as there is no guide on that, and I was unable to run them. I'll do it in the following days.

msbence:
I'm a bit lost here. I managed to write the tests (`TestAccProxmoxVmQemu_UpdateNoReboot` and `TestAccProxmoxVmQemu_UpdateRebootRequired`), but I don't know how to verify the results.

1. The `reboot_required` variable reset to `False` when reading the resource (`_resourceVmQemuRead`), so even if a reboot is needed, it will be `False` in the end of the change progress, and in the statefile.
2. Even if a change is "pending" (ie.: not applied) Proxmox-side, the API sends the desired values not the ones that are in effect, so I can not check if there are any not applied value.

Any ideas how can I write a test in this case?

mleone87:
You're right @msbence, it seems a bit tricky to test it with this kind of metodology, I think that would be sufficient to verify that make these changes do now reboot the vm(maybe checking the uptime?)

msbence:
Yes, that crossed my mind. But apart from there is no method for checking the uptime (at least none that I'm aware of), I think that checking the uptime is not reliable as it depends on the capabilities of the proxmox host. I'll look into this matter in the following days, but any suggestion is more than welcome!
Sidenote: one other thing is maybe we can check somehow PVEs task history. I have to look into the proxmox-go-api as I'm not that familiar with it, and I'm also not so sure how could I do that in a Terraform TestStep, but maybe that's one way to do it.

mleone87:
well, I think that we can agree that some "manual" testing is sufficient for now
working on the api project is a little difficult, maybe this could end in the "todo" list :)

msbence:
Given that always restarting VMs are a blocker for some of the people, I would agree that this should be merged despite the "not sufficient" testing. Anyways, I'll leave the two test cases here as testing the update process itself is a nice to have to say at least. :)
I'll look into the Go API and make an issue/PR there or here.

PR #293, opened 2021-01-18
Make finding VM's IP more robust

alexkrebiehl:
Adds a configurable timeout when trying to get the VM IP from the guest agent. The initial implementation was ~60 seconds, which is too short if part of the cloud-init is to install qemu-guest-agent (and other things).

When finding VM's IP, use MAC address of net0. This prevents interfaces added to the VM later (like Docker) from being matched.


alexkrebiehl:
@ggongaware any idea when this will be released?

PR #296, opened 2021-01-20
Fix links

LeMyst:


ggongaware:
Same problem as #260>

Changing the links will break the terraform registry view> https://registry.terraform.io/providers/Telmate/proxmox/latest/docs

PR #318, opened 2021-03-20
Fix broken links

vladistan:
The links from the documentation index.md to the resources pages do not work.   Here is a small patch to fix that.


msbence:
Same as #296 #260

vladistan:
I see...

PR #322, opened 2021-03-29
Fix import exist LXC rootfs crash

Kusumoto:
- Fix use `terraform import` command for import exist LXC has been crashed because cannot serialize `rootfs` config from API


ggongaware:
Nice fix, thank you

PR #348, opened 2021-05-07
Added the ability to debug the plugin through the Terraform Plugin SDK.

SourceFellows:
We encountered problems while using the plugin and we wanted to debug the code to see what's happening.

So we added, according to the documentation, the option to start the plugin in debugmode.

This patch is a simple implementation of the "Enabling Debugging In A Provider" documentaion here: https://www.terraform.io/docs/extend/debugging.html#enabling-debugging-in-a-provider

Thanks,
Kristian


tagur87:
@SourceFellows - You will need to rebase this now to follow the updated module design. 

Can you also add some documentation to the Readme on how to start the provider in debug mode? 

https://www.terraform.io/docs/extend/debugging.html#starting-a-provider-in-debug-mode

SourceFellows:
ok @tagur87, adjusted the Pull Request to the new layout.

PR #361, opened 2021-05-31
fix links in documentation index

OneMoreByte:
This pr fixes a minor annoyance I've run into when reading the documentation. When clicking the links for lxc and vm_qemu providers in the documentation index.md file, the url becomes 
```
https://github.com/Telmate/terraform-provider-proxmox/blob/master/docs/docs/resources/vm_qemu.md
```
instead of
```
https://github.com/Telmate/terraform-provider-proxmox/blob/master/docs/resources/vm_qemu.md
```
where the document is.


ggongaware:

PR #369, opened 2021-06-18
Fix incorrect path

jess-sys:
Line 4:
```
This provider exposes two resources: [proxmox_vm_qemu](docs/resources/vm_qemu.md) and [proxmox_lxc](docs/resources/lxc.md).
```

This path relative. Using docs/resources will lead to an absolute path like /docs/docs/resources, leading to a 404 error.

 I removed the unnecessary `docs/` from this line.


ggongaware:

PR #392, opened 2021-08-25
remove pointless variable

trioxtron:
Variable "waited" is not used, except for creating a while loop which is also possible by just creating a for loop without any argument in go -> removed "waited"


BaldFabi:
Could be merged. We already have a similar loop to wait until a vm is cloned

Syntax3rror404:
@BaldFabi But the existing logic doesn't work. Since 2.9.0 there is mostly a 500 error. Especially when you build several machines at the same time.

BaldFabi:
I see your point but the var doesn't get incremented which results in an infinite loop anyway.

mleone87:
this still misses a condition to break the loop in time, or at least before the proxmox hard timeout occours

PR #408, opened 2021-10-02
updated docs/resources/vm_qemu.md broken link

bryanpedini:
IMHO better than #389 because `/` could mean different things between GitHub and the `registry.terraform.io` documentation section website/framework.
But I might be totally wrong and YMMV, if so, feel free to accept #389 and just close this PR instead.


mleone87:

bryanpedini:
> Sorry this has the same problem as this PR #260

Mhh, even using relative links?

mleone87:
> > Sorry this has the same problem as this PR #260
> 
> Mhh, even using relative links? I even mentioned the registry documentation platform too... (tho I might be wrong on how that thing works)

relative links are ok but ../../docs is wrong anyway with the current docs structure, you can see that actual documentation is working fine on both gh and tf registry

Issue #411, opened 2021-10-06
interface conversion: interface {} is nil, not float64


pagaale:
Hi Team,
    This is my environment:
    plugin version: 2.8.0
    Proxmox web UI: 6.4-13
    I request a "proxmox_vm_qemu" resource with clone option in another pool, the machine is created successfully on proxmox (in the correct pool), after the plugin crashes with the error:
```
╷
│ Error: Plugin did not respond
│
│   with proxmox_vm_qemu.example,
│   on main.tf line 23, in resource "proxmox_vm_qemu" "example":
│   23: resource "proxmox_vm_qemu" "example" {
│
│ The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call. The plugin logs may contain more details.
╵

Stack trace from the terraform-provider-proxmox_v2.8.0 plugin:

panic: interface conversion: interface {} is nil, not float64

goroutine 28 [running]:
github.com/Telmate/terraform-provider-proxmox/proxmox._resourceVmQemuRead(0xc00019d180, 0xb7cf00, 0xc0001482d0, 0xc00024acd0, 0xc000571408)
        github.com/Telmate/terraform-provider-proxmox/proxmox/resource_vm_qemu.go:1423 +0x2eb9
github.com/Telmate/terraform-provider-proxmox/proxmox.resourceVmQemuCreate(0xc00019d180, 0xb7cf00, 0xc0001482d0, 0x0, 0x0)
        github.com/Telmate/terraform-provider-proxmox/proxmox/resource_vm_qemu.go:963 +0x1ed3
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).create(0xc00037a460, 0xda6e58, 0xc0003f41c0, 0xc00019d180, 0xb7cf00, 0xc0001482d0, 0x0, 0x0, 0x0)
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.6.1/helper/schema/resource.go:318 +0x1ee
github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema.(*Resource).Apply(0xc00037a460, 0xda6e58, 0xc0003f41c0, 0xc0003343f0, 0xc0004ccd40, 0xb7cf00, 0xc0001482d0, 0x0, 0x0, 0x0, ...)
        github.com/hashicorp/terraform-plugin-sdk/v2@v2.6.1/helper/schema/resource.go:456 +0x67b
[...]

Issue #388, opened 2021-08-20
The plugin encountered an error, and failed to respond to the plugin.(*GRPCProvider).ApplyResourceChange call.


radosevic:
Greetings,
I am bumping my head for days... Am I doing something wrong?
Every version ( tried 2.6.5, 2.6.6, 2.6.7, 2.7.1, 2.7.2, 2.7.3 and 2.7.4 ) of telmate/proxmox provider I tried gives me this panic/error:

`panic: interface conversion: interface {} is nil, not string`

Version of Proxmox:
```
root@some-proxmox:~# pveversion 
pve-manager/6.3-6/2184247e (running kernel: 4.19.0-16-amd64)
```
VM template is Ubuntu cloud image prepared by recommendations from Proxmox wiki + added qemu agent.
I get new cloned image, but provider breaks when starts doing disk block, so at the end disk is not resized and doesn't push cloud-init settings.

This is last entry in Proxmox-provider.log:
```
{"data":{"type":"dir","active":1,"total":391120543744,"avail":359123881984,"content":"iso,rootdir,vztmpl,images,snippets","used":24010919936,"shared":0,"enabled":1}}
unusedDiskNames: []
[90mFri, 20 Aug 2021 18:21:37 +0200[0m [33mDBG[0m [1mgithub.com/Telmate/terraform-provider-proxmox/proxmox/resource_vm_qemu.go:843[0m[36m >[0m Original disks: 'map[0:map[backup:1 cache:none discard:on file: format:qcow2 iothread:1 mbps:0 mbps_rd:0 mbps_rd_max:0 mbps_wr:0 mbps_wr_max:0 media:disk replicate:1 size:50G slot:0 ssd:0 storage:local storage_type: type:scsi volume:]]', Clone Disks 'map[0:map[file:1001/vm-1001-disk-0.qcow2 format:qcow2 size:2252M slot:0 storage:local storage_type:dir type:scsi volume:local:1001/vm-1001-disk-0.qcow2]]' [36mloggerName=[0mresource_vm_create [36mvmid=[0m
```

I can provide you with Terraform debug/trace logs if needed.

It is always breaking when doing disk block:
```
goroutine 97 [running]:
github.com/Telmate/proxmox-api-go/proxmox.(*Client).createVMDisks(0xc000364280, 0xc00011e6f0, 0x14, 0xc00033a7d8, 0x1d, 0x0, 0x0, 0xc000038d94, 0x7)
[...]

PR #422, opened 2021-10-22
Bugfixes, timeouts, logging improvement, some logic rework

mleone87:
Hi

I'm planning to do some more work on the provider but this is the first PR of the series

Some info:

- Made some additions to makefile to help local development
- Added **pm_debug** provider option (**PM_TIMEOUT** env variable) to set globally debug mode for the proxmox-api(it was set in every resource before now)
- Set **pm_tls_insecure** option to default to true since we assume it's a lab, production well configured environments should set to true
- Introduced the support to Terraform timeouts instead of using provider options or resource options. The new file is **timeouts.go**. Now it easy to get the timeouts in the resource lifecycle. More info there [Terraform Docs](https://www.terraform.io/docs/extend/resources/retries-and-customizable-timeouts.html)
- Moved some functions to **utils.go**
- Fixed some bugs, at least #411 #388 and I think countless more since a lot of issues are redundant
- Removed some random waits added to the provider in the past, more work needed on api side
- Fixed the docs and added the changes
- Some more I'm forgotting


TheUnixRoot:
I'm still getting the vm locked error after 2:40 minutes of creation, I think it is a timeout as cloning is slow

shdwlkr:
> I'm still getting the vm locked error after 2:40 minutes of creation, I think it is a timeout as cloning is slow

Me too. This happens if the template has more that one disk. When pm finished copying the first one - I got the "vm locked error".

Issue #435, opened 2021-11-06
conf to not auto start vm?


tinsel-grumble:
not seen in docs https://registry.terraform.io/providers/Telmate/proxmox/latest/docs/resources/vm_qemu how do keep vm power off after create as clone? vm seem to always auto power on, but setting to keep off is there when manual create

PR #436, opened 2021-11-09
Add option to not start vm on create

tuesdaycodes:
Resolves https://github.com/Telmate/terraform-provider-proxmox/issues/435

default is still to power on when vm is created so no backwards breaking change. users should see no difference with new option.

if set to false the vm will not be powered on after it is created. followed naming on `onboot` and updated docs with new option.

maintainer can edit if needed for new name or docs explanation. if tests are needed i might need help, couldnt figure out how to write them, i am new to go


tuesdaycodes:
@ggongaware can i plz have some help with the tests on this? i can do the build fine, and manual run works, but not good enough with go to get the tests passing.

proxmox/resource_vm_qemu.go
```diff
@@ -897,10 +903,19 @@ func resourceVmQemuCreate(d *schema.ResourceData, meta interface{}) error {
 	// give sometime to proxmox to catchup
 	//time.Sleep(time.Duration(d.Get("additional_wait").(int)) * time.Second)
 
-	log.Printf("[DEBUG][QemuVmCreate] starting VM: %d", vmr.VmId())
-	_, err := client.StartVm(vmr)
-	if err != nil {
-		return err
+	if d.Get("oncreate").(bool) {
+		log.Print("[DEBUG][QemuVmCreate] starting VM")
+		_, err := client.StartVm(vmr)
```
ggongaware, line 908: Careful with the scope of err, you'll see some build/test errors now

tuesdaycodes: i was able to build it locally. do i need to use = instead of := there for err, or does something else need moved around?

proxmox/resource_vm_qemu.go
```diff
@@ -897,10 +903,19 @@ func resourceVmQemuCreate(d *schema.ResourceData, meta interface{}) error {
...
+
+		err = initConnInfo(d, pconf, client, vmr, &config, lock)
+		if err != nil {
+			return err
+		}
+	} else {
+		log.Print("[DEBUG][QemuVmCreate] oncreate = false, not starting VM")
 	}
 
 	err = initConnInfo(d, pconf, client, vmr, &config, lock)
```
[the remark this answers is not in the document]
tuesdaycodes: fixed

Issue #445, opened 2021-11-18
Creating lxc container: "error 500 no such resource"


Zapfmeister:
Im trying to create an lxc container on a proxmox 7.1-5:

```
terraform {
  required_providers {
    proxmox = {
      source  = "telmate/proxmox"
      version = ">=2.9.1"
    }
  }
}

provider "proxmox" {
  pm_tls_insecure = true
  pm_api_url      = "https://myip:8006/api2/json"
  pm_password     = "password"
  pm_user         = "terraform-prov@pve"
}

resource "proxmox_lxc" "lxc-test" {
  count = 1
  features {
    nesting = true
  }

  // Terraform will crash without rootfs defined
  rootfs {
    storage = "local-zfs"
    size    = "8G"
  }

  hostname = "tf-new"
  network {
    name   = "eth0"
    bridge = "vmbr0"
    ip     = "dhcp"
    ip6    = "dhcp"
  }
  ostemplate = local:vztmpl/ubuntu-21.04-standard_21.04-1_amd64.tar.gz
  password   = "rootroot"
  target_node  = "host1"
  unprivileged = true
}
```

The container gets created, but it seems it can't get the containers state?:
```
proxmox_lxc.lxc-test[0]: Creating...
proxmox_lxc.lxc-test[0]: Still creating... [10s elapsed]
proxmox_lxc.lxc-test[0]: Still creating... [20s elapsed]
╷
│ Error: 500 no such resource 'ct:111'
│
│   with proxmox_lxc.lxc-test[0],
│   on main.tf line 20, in resource "proxmox_lxc" "lxc-test":
│   20: resource "proxmox_lxc" "lxc-test" {
│
```

I tried also with using root@pam, but the same issue appears.

Zapfmeister:
The ID 111 is correct

niko-lay:
same issue with 2.9.1 provider version

XDjackieXD:
I have the same issue (updated to 2.9.1 from 2.7.4) with all existing containers meaning I can't even get `terraform plan` to work because of the existing containers

NULLx76:
I've also got the same issue with provider version 2.9.1 after upgrading from 2.9.0, the containers with the container id do definitely exist.

gravyflex:
this worked for me after downgrading to 2.8.0.

mleone87:
a fix is waiting to be merged in api repo and then there

ebiekoelle:
[...]

Issue #443, opened 2021-11-16
Ip is not updated when applying


RemiDesgrange:
Terraform v1.05
proxmox-plugin v2.9.0

## Step to reproduce

* create a qemu VM which gets it's ip from dhcp
* VM ip changes
* Re terraform apply (without any changes to the `.tf` file
* IP is not updated in the terraform state.

123quhiwiwk:
Hi, I have the same issue (using Terraform v1.0.11 and Proxmox Provider 2.9.1).
I enabled logging and checked the HTTP requests/responses for the new IP address, but found nothing.
Normally the ip address comes from API `/api2/json/nodes/*name*/qemu/*id*/agent/network-get-interfaces`, but this API endpoint is not called when plan/reapply.
So I assume that this feature is not yet implemented.

Issue #440, opened 2021-11-11
[BUG] clone_wait deprecated now get 500 error while creating multiple vms


Syntax3rror404:
Hello,

i used the clone_wait argument to fix 500 error when creating more then one machine at same time. Since this argument isn't use anymore i get errors. 

Please reimport this argument to fix this error.

Error msg:

```
STDOUT:

module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[0]: Creating...
module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[1]: Creating...
module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[0]: Still creating... [10s elapsed]
module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[1]: Still creating... [10s elapsed]
module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[1]: Still creating... [20s elapsed]
module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[0]: Still creating... [20s elapsed]
module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[0]: Still creating... [30s elapsed]
module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[1]: Still creating... [30s elapsed]

Warning: Argument is deprecated

  with module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[1],
  on .terraform/modules/patrick_31_pve/vm.tf line 8, in resource "proxmox_vm_qemu" "proxmox_vm":
   8:   clone_wait        = 45

do not use anymore

(and one more similar warning elsewhere)



STDERR:


Error: 500 command '/usr/bin/qemu-img resize -f raw /mnt/pve/datastore/images/117/vm-117-disk-0.raw 68719476736' failed: got timeout

  with module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[1],
  on .terraform/modules/patrick_31_pve/vm.tf line 1, in resource "proxmox_vm_qemu" "proxmox_vm":
   1: resource "proxmox_vm_qemu" "proxmox_vm" {


Error: 500 command '/usr/bin/qemu-img resize -f raw /mnt/pve/datastore/images/110/vm-110-disk-0.raw 68719476736' failed: got timeout

  with module.patrick_31_pve.proxmox_vm_qemu.proxmox_vm[0],
[...]

Issue #437, opened 2021-11-10
Timeout clone


Gskill75:
Hello,
when I run a clone with template cloud-init I have this error:
`Error: 500 command '/usr/bin/qemu-img resize -f raw /var/lib/vz/images/113/vm-113-disk-0.raw 21474836480' failed: got timeout
 `
Have you ever had this problem? 
Can we change the timeout duration?

thanks

wOvAN:
2.9.0 got the same, 
pm_timeout - not helping

Syntax3rror404:
I also get this error. Befor this version i fixed this by using  clone_wait. But this argument is now deprecated :(

mleone87:
please @ggongaware push 2.9.1 version

Gskill75:
Hello
can you push the 2.9.1 quickly @ggongaware.
this problem present in previous version?

thanks

ggongaware:
Building now...

Syntax3rror404:
Issue still existing on 2.9.1

mleone87:
@Syntax3rror404 @Gskill75 timeouts for resource creation AND api operations is now 5 minutes by default, if your operation is longer update both pm_timeout and the resource creation limits like there

https://www.terraform.io/docs/language/resources/syntax.html#operation-timeouts

Gskill75:
Hello  @mleone87,
the pm_timeout not working for me.

mleone87:
@Gskill75 provide detailed info and logs

Gskill75:
provider file:
`
`terraform
terraform {
  required_version = "~> 1.0"
    backend "s3" {
      bucket  = "scms-s3-terraform-remote-state-storage"
      key     = "terraform.tfstate"
      region  = "eu-west-1"
      encrypt = true
    }
  required_providers {
    aws = {
      source = "hashicorp/aws"
      version = "3.63.0"
    }
    proxmox = {
      source = "Telmate/proxmox"
      version = "2.9.1"
    }
  }
}

provider "proxmox" {
    pm_api_url      = "https://x.x.x.x:8006/api2/json"
    pm_tls_insecure = "true"
    pm_timeout      = 1800 
}
provider "aws" {
  region = "eu-west-1"
}
```
`

`resource "proxmox_vm_qemu" "scms_prod_dock-01" {
    name                   = "scms.prod.dock-01"
    target_node            = "x"
    clone                  = "template-deb10"
    full_clone             = "true"
[...]

PR #451, opened 2021-11-23
fix for some issues occurred

mleone87:


mleone87:
fixes #445 , #443 , #440 , #437 , #431, #426 , #424

PR #463, opened 2021-12-09
Add new `aio` disk attribute and endorce `scsihw` values

RemiDesgrange:
I added the `aio` disk attribute, which was missing. And enforce the `scsihw` value. I'm not a go developer, so I took the `StringInSlice` method from this: https://github.com/hashicorp/terraform-provider-aws/issues/14601.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -195,6 +195,14 @@ func resourceVmQemu() *schema.Resource {
 				Type:     schema.TypeString,
 				Optional: true,
 				Computed: true,
+				ValidateFunc: validation.StringInSlice([]string{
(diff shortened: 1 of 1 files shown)


Syntax3rror404:
@RemiDesgrange thank you for your fix :) which version do you use? I think 2.8.0? The versions after 2.9.0 are buggy because you get "VM already running" error or if you run tf parallel existing machines will be overwritten

RemiDesgrange:
I'll check that. I'll update the code of the `resourceVmQemuUpdate` to make the upgrade working.

Syntax3rror404:
Its a general 2.9 and newer problem. The only working version is 2.8.0. Thats the sad story :(

PR #469, opened 2021-12-18
vm deletion graceful shutdown

H4R0:
Currently scaling vm's down will kill them. This is really bad since it can easily result in data loss.

I use this method to reinitiate the whole cluster / update it using IaC.

This pr will try a graceful shutdown first, like it is handled with QemuVmUpdate and fallback to stop.

Note: if qemu guest agent is not installed in the vm but enabled in terraform this will result in a freeze of 300 seconds, sadly this has not been patched in proxmox upstream.

Instead we could also add a resource option to enable graceful shutdowns manually but IMHO this should be the default.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -110,6 +110,7 @@ The following arguments are supported in the provider block:
 - `pm_timeout` - (Optional; defaults to 300) Timeout value (seconds) for proxmox API calls.
 - `pm_debug` - (Optional; defaults to false) Enable verbose output in proxmox-api-go
 - `pm_proxy_server` - (Optional; defaults to nil) Send provider api call to a proxy server for easy debugging
+- `pm_delete_with_graceful_shutdown` - (Optional; defaults to false) Use graceful shutdown instead of hard stop on vm/lxc deletion for down scaling
 
 Additionally, one can set the `PM_OTP_PROMPT` environment variable to prompt for OTP 2FA code (if required).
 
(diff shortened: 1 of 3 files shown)


mleone87:
sorry, i miss the point in gracefully shut down a vm that will be deleted

H4R0:
> sorry, i miss the point in gracefully shut down a vm that will be deleted

if it uses local file storage only yes in theory it makes no sense.

But with IaC/Kubernetes and many other setups you mount network storage as persistent storage.

Terraform is exactly designed for this and graceful shutdown on delete is done in other providers like ESXI/vSphere, so I would expect it to be the same with Proxmox.

H4R0:
> I am also of the opinion that a gracefully shut down makes no sense. It makes the deletion process slower and that for a VM that you delete anyway. 

It takes like 10-20 seconds I don't see how this is a problem.

> I also run Rancher RKE2 with this Terraform provider. And the step is to remove the Kubernetes cluster before deleting and then to execute the destroy.

Same goes for down scaling e. g. 5 to 3 nodes. In that case you don't delete the whole cluster and k8s automatically redistributes the workload. There shouldn't be any other mandatory steps needed with Terraform apply.


I'm fine with introducing a provider option for example:
```
provider "proxmox" {
  # reducing resource count will do a graceful shutdown 
  # of running instances instead of hard stop
  scale_with_graceful_shutdown = true
}
```
But this should still considered to be the default (in next minor/major) or at least we need to add a warning in the docs that data loss is likely to happen with down scaling.

H4R0:
added provider option which defaults to old behavior

PR #476, opened 2021-12-28
Add automatic_reboot parameter to proxmox_vm_qemu

c0deaddict:
VM's are currently automatically rebooted when certain parameters are changed. We are running quite a few production VM's based on a Terraform config with this provider and don't want those VM's to get rebooted automatically when anything changes.

This PR introduces a new parameter `automatic_reboot`. If the parameter is set to `true` (default value) then the old behaviour remains. If set to `false` the provider will not reboot the VM but instead will emit a warning message that a reboot should be done manually (i.e. on a later scheduled time).


mleone87:
Thanks! It would be also nice to migrate all the resource schema to the new sdkv2 *Context like you do with Update

PR #495, opened 2022-02-05
Support PXE for install usecase

USA-RedDragon:
Currently, the PXE usecase does not allow for using PXE for provisioning and booting from a disk, as the PXE mode requires network to be the first option.

By setting the boot order to the hard drive first and then network will allow hosts to be provisioned on the first boot and run from disk on subsequent boots.

This might not be the most optimal way of dealing with this usecase. I'll be happy to help implement it correctly if changes are required.


mleone87:
@sygibson

sygibson:
@USA-RedDragon - typically in all of my PXE server setups, I use the nextboot directive to determine whether to boot off a network installer or tell the machine to boot to local disk.  This enables full control from a single point of management across 10s of thousands of hosts.  This ensures that the machine **always** PXE boots, and the boot path is controlled from one service instead of all of the disparate machines random boot config settings, which are often ... varied.

Once the Machine has been PXE booted and installed - it's certainly feasible to change the boot order of the Machine to boot local disk first if desired.  This should be independent of the Terraform use case that is creating a new machine - in which case you want to guarantee that the machine boots to PXE first.  

I have seen installations that use existing "disk objects" to back the VM, which may or may not have an existing OS on it.  In those cases, you do not want the disk to boot to local - and you'd never gain control over the machine to enable a new provisioning operaiton.

**_Do you have a suggestion on how you'd like to see the behavior?_**  I'm open to understanding your use case, and how we can best adjust the code to support both uses.

I could add an `pxe-always` option (true/false), which if set to true would enable the `boot` argument verification code.  If it were set to `false`, then no boot arguments would be verified, and the operator would be free to specify any boot settings without any validations/enforcement.

gunzy83:
Any chance this can be merged? I have run into this as well as this static boot order install use case works well across many machines with or without IPMI/management interfaces and VMs (in proxmox). I am even playing with a deprovisioning workflow using a destroy provisioners to drain the node then render the install drive non-bootable and return to the PXE boot on next provision.

mleone87:
are we ok with this PR @USA-RedDragon  @sygibson, do we need other commits?

sygibson:
Yes, I'm ok with this change.

mleone87:
will be in the next release train 👍

shantanugadgil:
I would concur with @sygibson here ...

I had a created a PXE installation automation (hands off) where the `network boot` was always the first option on the client and the default option on the PXE server was of `localboot 0` (first HDD). It had options of second hdd, third hdd, etc, just for the sake of completeness.

When a client machine was targeted for a reinstall, the machine's MAC id specific PXE file would be created on the PXE server, with ONLY the OS installation as the first option.
This "machine specific" file would be removed when the installation starts (i.e. pre install step)  or after a timeout ~ 30 minutes.

_Any recent machine completes a disk format + OS install in less than 10-12 minutes_

gunzy83:
@shantanugadgil Thanks for sharing your use case. If it wasn't clear the PR does not remove the original functionality, just allows for a less restrictive config that allows for my and Jacob's use cases.

PR #499, opened 2022-02-14
Fix MTU type for LXC container

vmvarga:
Currently mtu for LXC container is expected as string and there is no possibility to use this option.


VitalyZnachenok (commented):
Duplicated with https://github.com/Telmate/terraform-provider-proxmox/pull/498

vmvarga:
> Duplicated with #498

Please add modifications to resource_lxc.go because your modification is only for VM.

PR #545, opened 2022-04-24
Start parameter was not taken into account when clone of LXC was done

### LINKED ISSUE #515: LXC with clone did not modify some flags ever
`start ` and `onboot ` flags did not works with `clone`

```tf
 resource "proxmox_lxc" "worker1" {
   target_node = "node1"
   hostname = "worker1"
   clone = 100
   full = true
   memory = 1024
   start = true
   onboot = true
 }
```
After first `terrform apply` CT  will be created  with `start = false`
After second`terrform apply` CT will be trying to `start = false` -> `start = true` but CT will be stopped anyway (`start = false`).

vmvarga:
This should fix #515

### DIFF
--- a/proxmox/resource_lxc.go
+++ b/proxmox/resource_lxc.go
@@ -573,6 +573,18 @@ func resourceLxcCreate(d *schema.ResourceData, meta interface{}) error {
 		if err != nil {
 			return err
 		}
+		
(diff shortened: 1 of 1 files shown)


mleone87:
update resource code is missing

Issue #525, opened 2022-03-27
Is there an argument for QEMU VM start up order?


ralgar:
I see there is an argument `startup` for LXC, however I cannot seem to find any related argument for VMs in the docs.

I've tried using the same `startup` argument for VMs and Terraform returns `Unsupported argument`.

BaldFabi:
Have a look at https://github.com/Telmate/terraform-provider-proxmox/blob/master/docs/resources/vm_qemu.md#top-level-block for the option `onboot`

ralgar:
Hi @BaldFabi,

I did find the `onboot` option, however that is only a boolean whether to start the VM or not, there seems to be no option to set the startup order and delay, as there is with LXC.

You can [see here](https://github.com/Telmate/terraform-provider-proxmox/blob/master/docs/resources/lxc.md#optional) the option which I am talking about is called `startup`, and it is supplementary to the `onboot` option.

BaldFabi:
Yes you're right. I misinterpreted the documention. As far as I know it's currently impossible. It could be implemented if the api provides it.

ralgar:
It seems that the option is there for both POST and PUT, according to the [API docs](https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/config).

I'm currently experimenting with this locally. I have no Go experience, but if I'm not mistaken it seems like only a few lines of code to implement it. If I can get it working, I'll submit a PR sometime this week.

PR #570, opened 2022-06-23
Add missing QEMU startup order option

ralgar:
This PR is for #525, and implements the currently missing startup order option for QEMU VMs.

There is a prerequisite for this PR at https://github.com/Telmate/proxmox-api-go/pull/182, which adds the required code in the API module. Of course, this will also require updating the imports in this repository, which is not included in this PR. I assume the maintainers would handle this chore on the next versioned release.

The code is not organized alphabetically in this module, so I put the added code in what seemed to be the most logical place, next to `onboot`. Hopefully this is OK, but if not, feel free to make edits and move things around as needed.

Everything is tested and working as expected on my end. The only exception is that the startup option cannot be nullified once it is set. As far as I can tell, this is a limitation of the Proxmox API.


mleone87 (approved):

mrtwnklr:
Is a new release already in sight?
I am looking forward to use the `startup` option in my configuration!

Issue #284, opened 2021-01-05
Unused disk is created on change unless I specify volume and file.


asenmx:
Hello, I've noticed that when i change something new unused disk is created and used as "boot" which breaks my vm.

```
resource "proxmox_vm_qemu" "dc0test" {
  provider = "proxmox.dc0"
  target_node = "dc0n5"
  vmid = "8407"
  clone             = "Ubuntu-18.04LTS-wCI-4-tpl"
  full_clone      = true
  name              = "vm-1-tf-test"
  cores             = 2

  bootdisk = "scsi0"
  disk {
    size            = "5G"
    type            = "scsi"
    storage         = "vm-ssd"
    file = "vm-8407-disk-0"
    volume = "vm-ssd:vm-8407-disk-0"

  }
}
```

If I do it like this it works, however if I remove the following 2 lines it creates new disk and puts it as boot drive.

```
    file = "vm-8407-disk-0"
    volume = "vm-ssd:vm-8407-disk-0"

```

Terraform v0.14.3
Installed telmate/proxmox v2.6.6

mleone87:
@dabicata  why do you specify volume an file? it's redundant

also, removing them should create a new disk as you remove all the references to the working disk

dabicata:
@mleone87 I want the first time created disk after the cloning to stay, but if I don't specify the volume and file new disk is created every time i apply the terraform.

mleone87:
mm that's quite unusual and breaks the purpose of terraform

can you post the resource definition that creates a new disk every time you apply it?

dabicata:
> mm that's quite unusual and breaks the purpose of terraform
> 
> can you post the resource definition that creates a new disk every time you apply it?

sure just remove the file and volume from the one i posted like this

resource "proxmox_vm_qemu" "dc0test" {
  provider = "proxmox.dc0"
  target_node = "dc0n5"
  vmid = "8407"
  clone             = "Ubuntu-18.04LTS-wCI-4-tpl"
  full_clone      = true
  name              = "vm-1-tf-test"
  cores             = 2

  bootdisk = "scsi0"
  disk {
    size            = "5G"
    type            = "scsi"
    storage         = "vm-ssd"

  }
}

yukron:
That's definitely what is happening, new unused disk is created all the time if not strictly specified.

mleone87:
@yukron 

can you post your terraform resource definition and you clone template configuration?

yukron:
[...]

PR #582, opened 2022-07-26
Backup should be a Int

DoktorAerzt:
This PR ensure that backups is a int and not a bool. See
```
[ERROR] setting state: disk.1.backup: '' expected type 'int', got unconvertible type 'bool', value: 'true'
```

With this PR #284 should be fixed.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -749,7 +749,7 @@ func resourceVmQemuCreate(d *schema.ResourceData, meta interface{}) error {
 		Pool:         d.Get("pool").(string),
 		Bios:         d.Get("bios").(string),
 		Onboot:       d.Get("onboot").(bool),
-		Startup:	  d.Get("startup").(string),
+		Startup:      d.Get("startup").(string),
 		Tablet:       d.Get("tablet").(bool),
 		Boot:         d.Get("boot").(string),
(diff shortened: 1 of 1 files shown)


mleone87:
this would make things worse, did you test your code?

DoktorAerzt:
Yes i tested the code. 

When you use bool for backup than you get 
```
[ERROR] setting state: disk.1.backup: '' expected type 'int', got unconvertible type 'bool', value: 'true'
```
and also you are missing the volume and file entry's in the state file of your vm. 
When you change something then terraform assumes, that your vm disk has no name and will create a new one. 

When you use int for backup than the error msg is gone and in your state file of your vm the entry's for volume and file are present and you can change some values after deployment and terraform will not recreated your disks.

DoktorAerzt:
The fix is necessary because in line https://github.com/Telmate/terraform-provider-proxmox/blob/master/proxmox/resource_vm_qemu.go#L353 you set backup to be a int if i'm correct.

DoktorAerzt:
It was changed in this pr https://github.com/Telmate/terraform-provider-proxmox/pull/287 and there the change what i made is missing. So after that pr you want to write an bool into an int. 

This pr should also fix https://github.com/Telmate/terraform-provider-proxmox/issues/283

Issue #605, opened 2022-08-30
Proxmox API has new `import-from` parameter


jamesharr:
I'd like to add the `import-from` parameter for the disk block. I think this would be very useful.

Currently this terraform provider allows you to create a VM by using a clone, or specifying an ISO. One of the two are required. However there is a third option - most distributions create a qcow2 disk image that can be imported into a hypervizor. Disk images typically support cloud-init, perform system updates on first boot, and while the image is sized for a 10GB disk, it will grow to fit available size automatically.

Proxmox-VE now has the ability to import disks on VM create. An `import-from` parameter can be used on `diskN` to specify the disk image. This disk image is most commonly a qcow2 and the format is independent of the datastore the VM disk will be stored on. This is really nice, for example, if you're using an OS that has a pre-built "cloud" image that has cloud-init in it already.

References:
- [Proxmox VE API reference](https://pve.proxmox.com/pve-docs/api-viewer/index.html#/nodes/{node}/qemu/{vmid}/config)
- [Proxmox VE Roadmap - 7.2](https://pve.proxmox.com/wiki/Roadmap#Proxmox_VE_7.2) - search for "importing disk-images"

Example pre-built images
   - [AlmaLinux 9](http://ord.mirror.rackspace.com/almalinux/9.0/cloud/x86_64/images/)
   - [Ubuntu 22.04](https://cloud-images.ubuntu.com/releases/22.04/release/) - look for the `-disk-kvm.img` variant
   - [Debian bullseye](https://cdimage.debian.org/images/cloud/bullseye/)

### Note on multi-disk VMs
Below are some notes on our specific use-case. I believe they may apply to others, however that is up for discussion.

In some cases, we actually need more than one disk. The first disk would be the system disk (the cloud-init image), typically sized 50-100GB, but we also are going to use a second or third disk that's 300-1000GB that's used for data storage. This is very similar to something like adding an extra EBS volume to an AWS EC2 instance.

I'm not sure I can imagine a use-case where we'd want to import multiple volumes, however I'm not sure I'd rule it out either. If this TF provider has an opinionated design (that's ok if it does), then a reasonable compromise might be specifying the `import-from` parameter at the QEMU block and then it's simply applied to the first disk.

The two trade-offs I see to specifying import-from at the QEMU block instead of at the Disk block:
- `import-from` only applying to the first disk is implicit behavior instead of explicit.
- It only allows a single disk import, which is probably practical.

### Note on disk paths
[...]

PR #606, opened 2022-08-30
feat: Add 'import-from' parameter to qemu

jamesharr:
I think this addresses #605. It adds an optional `import-disk` parameter for VMs.

I have no idea how to test this, or if I'm getting the additions made in the correct place. Please review with scrutiny and/or modify as-needed.


mleone87:
This feature should be used in the next version of the provider for sure!

However, I feel that putting this in the resource properties would cause a lot of problem, this is not a parameter to be passed to the API every time we make a change(just an example).
The proxmox developers thought of it as a one-time pass parameter so it should be used only in resource creation(or recreation)

jamesharr:
Ty. I agree with that assessment. For now, we're cloning from a template VM, but I would like to use this in the future.

I might need some pointers on how to accomplish this... I have limited experience using this TF provider let alone developing a TF provider, so I'm kind of lost with how to only apply that parameter on initial creation. I'll poke around today a bit but I'm not sure how far I'll get.

One other thing I noticed is that this might be an alternative to using iso or clone on the VM. When importing a disk image, neither parameter is needed. At the moment, one of those options is required, so I think this patch needs to add something to allow for a 3rd option instead of import/iso.

jamesharr:
I have a local development environment going. I found another issue that we'll run into, specifically that I want to use cloud-init without cloning from an existing template.

```
proxmox_vm_qemu.control-node: Creating...
╷
│ Error: cloud-init parameters only supported on clones or updates
│ 
│   with proxmox_vm_qemu.control-node,
│   on main.tf line 93, in resource "proxmox_vm_qemu" "control-node":
│   93: resource "proxmox_vm_qemu" "control-node" {
│ 
╵
Command failed 1
```

I'm going to modify the PR text to include more details of my proposed use-case since this is turning into a bigger thing than just adding a parameter. That way, I'll have something to work from.

myplixa:
Hi guys, can you tell me when will be the release of this feature?

TomyLobo:
@Tinyblargon this change didn't survive the transition to 3.x.
It was removed in https://github.com/Telmate/terraform-provider-proxmox/pull/926/commits/5b89c2ea9c23b7dfbbcb047157f6679b1631d656#diff-104365919f693375882979581d9f36f5266991eef361948e56c220931fe886ddL600 and never re-added.

Is it possible to add this back?
It's really useful in order to import official disk images.

PR #610, opened 2022-09-09
Check if CloudInit Ip is already present on the vm and sets the sshHost to the cloudInit Ip.

### LINKED ISSUE #593: CloudInit IP Should override default_ipv4_address and ssh_host
At the moment when you use cloudinit, the default_ipv4_address and ssh_host gets set on the temporary dhcp address witch gets overriden when the vm applys the cloudinit file. 

I think https://github.com/Telmate/terraform-provider-proxmox/blob/master/proxmox/resource_vm_qemu.go#L1865 should be called after this https://github.com/Telmate/terraform-provider-proxmox/blob/master/proxmox/resource_vm_qemu.go#L1892 block and the if https://github.com/Telmate/terraform-provider-proxmox/blob/master/proxmox/resource_vm_qemu.go#L1869 needs to be removed or we need to search a other way. 

What are your opinions, should i make a pr?

DoktorAerzt:
Fixes #593

This Pull Request checks if the vm has already the cloud init ip and if so then the sshHost and default_ipv4_address will be the cloudinit ip and not the dhcp ip address.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -1927,10 +1927,10 @@ func initConnInfo(
 	}
 	log.Print("[DEBUG][initConnInfo] trying to find IP address of first network card")
 
-	interfaces, err := client.GetVmAgentNetworkInterfaces(vmr)
 	// wait until we find a valid ipv4 address
 	for guestAgentRunning && time.Now().Before(guestAgentWaitEnd) {
 		log.Printf("[DEBUG][initConnInfo] checking network card...")
+		interfaces, err := client.GetVmAgentNetworkInterfaces(vmr)
 		net0MacAddress := macAddressRegex.FindString(vmConfig["net0"].(string))
 		if err != nil {
 			return err
@@ -1969,14 +1969,19 @@ func initConnInfo(
 					sshHost = ipMatch[1]
 				}
 				ipconfig0 := net.ParseIP(strings.Split(ipMatch[1], ":")[0])
-				for _, iface := range interfaces {
-					if sshHost == ipMatch[1] {
-						break
-					}
-					for _, addr := range iface.IPAddresses {
-						if addr.Equal(ipconfig0) {
-							sshHost = ipMatch[1]
+				interfaces, errInterfaces := client.GetVmAgentNetworkInterfaces(vmr)
+				if errInterfaces != nil {
+					return err
+				} else {
+					for _, iface := range interfaces {
+						if sshHost == ipMatch[1] {
 							break
+						}
+						for _, addr := range iface.IPAddresses {
+							if addr.Equal(ipconfig0) {
+								sshHost = ipMatch[1]
+								break
+							}
 						}
 					}
 				}
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -1971,7 +1971,7 @@ func initConnInfo(
 				ipconfig0 := net.ParseIP(strings.Split(ipMatch[1], ":")[0])
 				interfaces, errInterfaces := client.GetVmAgentNetworkInterfaces(vmr)
 				if errInterfaces != nil {
-					return err
+					return errInterfaces
 				} else {
 					for _, iface := range interfaces {
 						if sshHost == ipMatch[1] {
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -1877,7 +1877,9 @@ func initConnInfo(
 ) error {
 
 	var err error
+	var interr error
 	var lasterr error
+	var interfaces []pxapi.AgentNetworkInterface
 	// allow user to opt-out of setting the connection info for the resource
 	if !d.Get("define_connection_info").(bool) {
 		log.Printf("[DEBUG][initConnInfo] define_connection_info is %t, no further action\n", d.Get("define_connection_info").(bool))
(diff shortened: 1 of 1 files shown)


BryceTech122:
Any idea as to when this might be merged? Im trying to make a group of vms on one target_node in a cluster and it creates the vm's correctly every time but in the terraform.tfstate and when i try to run the below code block it always complains that the default_ipv4_address is null
`data "template_file" "k8s" {
  template = file("./templates/k8s.tpl")
  vars = {
    k3s_master_ip = "${join("\n", [for instance in proxmox_vm_qemu.proxmox_vm_master : join("", [instance.default_ipv4_address, " ansible_ssh_private_key_file=", var.pvt_key])])}"
    k3s_node_ip   = "${join("\n", [for instance in proxmox_vm_qemu.proxmox_vm_workers : join("", [instance.default_ipv4_address, " ansible_ssh_private_key_file=", var.pvt_key])])}"
  }
}`

DoktorAerzt:
> Any idea as to when this might be merged? Im trying to make a group of vms on one target_node in a cluster and it creates the vm's correctly every time but in the terraform.tfstate and when i try to run the below code block it always complains that the default_ipv4_address is null `data "template_file" "k8s" { template = file("./templates/k8s.tpl") vars = { k3s_master_ip = "${join("\n", [for instance in proxmox_vm_qemu.proxmox_vm_master : join("", [instance.default_ipv4_address, " ansible_ssh_private_key_file=", var.pvt_key])])}" k3s_node_ip = "${join("\n", [for instance in proxmox_vm_qemu.proxmox_vm_workers : join("", [instance.default_ipv4_address, " ansible_ssh_private_key_file=", var.pvt_key])])}" } }`

This PR shouldn't fix your problem. In your case its a race condition. This PR is only necessary, when you use cloudInit and DHCP in the same network. 
Then your VM is getting a DHCP address (witch is written to default_ipv4_address) and the check for the cloudInit address would never run and this PR changes it. 

Your problem is that your VM has no IP address at the moment the provider checks for all addresses. At first i would increase the terraform create timeout because this is just for the "IP Check Loop" here https://github.com/Telmate/terraform-provider-proxmox/pull/610/files#diff-104365919f693375882979581d9f36f5266991eef361948e56c220931fe886ddR1931 in my PR.

But that part is old code, so like i said not related to this PR.

proxmox/resource_vm_qemu.go
```diff
@@ -1926,11 +1927,11 @@ func initConnInfo(
 	}
 	log.Print("[DEBUG][initConnInfo] trying to find IP address of first network card")
 
+	interfaces, err := client.GetVmAgentNetworkInterfaces(vmr)
```
mleone87: if this fails the first time then it will break any other check later, easy to happen thought, I have a lot of vm failed in terraform but ok on proxmox because of that, especially in batch of 2+ vms

proxmox/resource_vm_qemu.go
```diff
@@ -1952,26 +1953,36 @@ func initConnInfo(
...
-					// parse IP address out of ipconfig0
-					ipMatch := rxIPconfig.FindStringSubmatch(d.Get("ipconfig0").(string))
+			if d.Get("ipconfig0").(string) != "ip=dhcp" || vmState["agent"] == nil || vmState["agent"].(float64) != 1 {
+				// parse IP address out of ipconfig0
+				ipMatch := rxIPconfig.FindStringSubmatch(d.Get("ipconfig0").(string))
+				if sshHost == "" {
 					sshHost = ipMatch[1]
 				}
+				ipconfig0 := net.ParseIP(strings.Split(ipMatch[1], ":")[0])
+				for _, iface := range interfaces {
```
mleone87: as per the previous comment, with interfaces = nill this will loop until the end of time

proxmox/resource_vm_qemu.go
DoktorAerzt: Alright, i moved it back to it old position and the old err != nil check handels it.

proxmox/resource_vm_qemu.go
DoktorAerzt: Copy the check for errInterface != nil
Or should i ruse the err variable at this place?

proxmox/resource_vm_qemu.go
mleone87: interfaces will be ok if the agent will come up at same point, so the code should work unless we reach the timeout limit

proxmox/resource_vm_qemu.go
mleone87: look at this schenario(like mine)
I make a vm with cloud init, the first part of the function will find the ip in some minutes and its' ok, but in my cloudinit conf i make the vm reboot after the install of qemu-guest-agent(mainly to reboot to the updated kernel after the system upgrade for example)
the ip should be ok but then with your PR we check it another time just to overwrite it with what is in the cloudinit conf
but in this situation the vm is rebooting so the check for the interfaces give us an error 500 and the code crashes because it's not handled
this will not happen 100% because it depends on how fast the process is, but with a lot of vm created at once(like a lot of people there like to do thinking they have AWS grade servers), a lot of random things can happen

DoktorAerzt: Ok, now i unterstand. 

So i moved interface to a "function variable" and in the "default" IP loop i rewrite the variable interface after this loop i just reuse it in the cloudinit entry. So when the first loop won't end in an infinity loop then the second can't end in one. 

Should i search for a other solution?

mleone87:
merged with some changes

PR #618, opened 2022-10-07
feat: add http_headers, add hostpci resource

BenediktBertsch:
Hey @mleone87,

I've tried to implement the new http_headers and also my previous PR for hostpci, but it seems like that adding hostpci devices is not working. In the tfstate is the hostpci existing, but not on the host.

created vm with hostpci:
image

tfstate:
```
            "hostpci": [
              {
                "host": "0000:01:00.1",
                "rombar": 0
              }
            ],
```

Using normal json with the proxmox api works, but not my implementation. Have I missed some other settings which have to be set?


mleone87:
let me check your PR locally and see! nice work btw!

mleone87:
@BenediktBertsch pull latest api and this now should work

BenediktBertsch:
@mleone87 thanks for fixing this issue in proxmox-api! Looks good and works locally.

brutesque:
I'm trying to implement this. Do you have an example configuration?

PR #668, opened 2023-01-23
Fix boot order option documentation, disallow deprecated syntax

tnyeanderson:
The syntax described in the documentation does not work (it has been deprecated). This fixes the regex check on the value so it will fail if the correct syntax isn't used, and updates the documentation to accurately reflect upstream.

The long version: using `acdn`-style boot order has been deprecated, and now you must use e.g. `legacy=acdn` or `order=device1;device2`

EDIT: `legacy=` doesn't work either, the whole thing needs deprecated. See my comment below.


tnyeanderson:
Nevermind, just tested with the `legacy=cdn` label and that doesn't work either. I think this is because it states in the documentation that only devices listed in `order=` will be marked as bootable. And if you look at the boot order in the GUI using `legacy=` the disk (`scsi0`) will not be checked.

Point of this is to support PXE-based installations. The answer is to use the equivalent `order=scsi0;ide2;net0` value, which does work, but isn't necessarily as flexible (you have to specify exact device names instead of "any disk, then any cd, then any network").

Since this basically completely breaks the legacy, deprecated notation, it probably shouldn't be supported in this provider either :)

Issue #429, opened 2021-10-28
Unable to disable disk backup


Tinyblargon:
setting the backup propertie to 0 does not unset the backup check mark in the proxmox gui.

```
  disk {
    type = "scsi"
    storage = "local-lvm"
    size = "2G"
    backup = "0"
  }
```

When unchecking the backup checkbox inside the proxmox gui the disk configuration is as followed:
`local-lvm:vm-9001-disk-0,backup=0,size=2G`

When setting `backup = "0"` in terraform it results in:
`local-lvm:vm-9001-disk-0,size=2G`
By default backup is turned on in proxmox and only shows in the disk configuration when it is disabled.
The checkbox in the gui is also still checked.
This leeds me to believe that setting `backup = "0"` in terraform did not turn the backup off.

Terraform Provider version = 2.9.0
Proxmox version = 7.0-13

n8rade:
Running into this same issue with the following versions:

PVE 7.2-7
Terraform 1.2.6
Telmate/terraform-provider-proxmox 2.9.10

There's a lot that's interesting about this.
1. The documentation states that the default value is 0 here. https://registry.terraform.io/providers/Telmate/proxmox/latest/docs/resources/vm_qemu#disk-block
Despite this being the default (and me missing that the default was to exclude backing up any disks), any new VM I create with this default value is included in any backups I perform.

2. All of the machines that I've imported into my terraform state all have "backup=0" under the disks section, despite all of them being included in backups before I started using Terraform. In other words, Proxmox sees the disks as being available to be backed up. None of them have a "backup=" value defined on the disk, but the box is check marked if I open up the disk to view in the UI.

3. Here are the strange behaviors I see when testing this:

### Test 1:
**State of Disk:** "backup=0" set manually on the disk
**Resource Expects:** "backup=0"
**Terraform Plan:** No changes expected

### Test 2:
**State of Disk:** "backup=0" set manually on the disk
**Resource Expects:** "backup=1"
**Terraform Plan:** Wants to change
**Terraform Apply:** Sets "backup=1" and also checks the "Backup" box in the Proxmox UI

### Configure seen in Proxmox UI from terraform for **Test 2**:
~~~
[...]

PR #669, opened 2023-01-24
change backup disk option to boolean

tnyeanderson:
In the [upstream docs](https://pve.proxmox.com/wiki/Manual:_qm.conf) the `backup` option is a boolean.

See reasoning in https://github.com/Telmate/proxmox-api-go/pull/231 for why the default value is true. That PR and this one should finally allow backups to be disabled.

Fixes: #429


tnyeanderson:
Looks like the container side is already correct.

>By default additional mount points besides the Root Disk mount point are not included in backups. For volume mount points you can set the Backup option to include the mount point in the backup. Device and bind mounts are never backed up as their content is managed outside the Proxmox VE storage library.

This should be tested the PR from `proxmox-api-go` and should be ready to merge.

PR #688, opened 2023-02-11
Add Missing Required Permission

scottbot95:
`Sys.Modify` permission is required in order to modify the `startup` value of a VM. Since this provide purports to support modifying the `startup` field, add `Sys.Modify` to the list of minimum permissions. Also fix several inconsistencies in the docs around setting up the PVE role.

(3rd PR in a week "fixing" this. turns out permissions are hard, who knew?)


mleone87:
what a mess, thanks! 👍

PR #690, opened 2023-02-11
Do not return error if MAC address value is empty

NlVlN:
Proxmox will generate a new MAC address when the provided value is empty and therefore MacAddressValidator function should not fail if the provided value is an empty string.


NlVlN:
Issue for #691

PR #696, opened 2023-02-19
Fix Ipconfigmap empty values

tnyeanderson:
See the following correlated PR for an issue description and details:

https://github.com/Telmate/proxmox-api-go/pull/238

Currently, every `ipconfigX` config property is set to an empty string no matter what. This is due to the use of `.Get()`:

>Get returns the data for the given key, or nil if the key doesn't exist
in the schema.
>
>If the key does exist in the schema but doesn't exist in the configuration,
then the default value for that type will be returned. For strings, this is
"", for numbers it is 0, etc.

The problem is this will always return an empty string. That mixed with the `.(string)` cast makes for a bug.

The map value for each `ipconfig` entry should not be set unless it actually is set in the config. This PR uses `.GetOk()` instead, and only sets the map value when there actually is one.


tnyeanderson:
@mleone87 gentle ping for this and https://github.com/Telmate/proxmox-api-go/pull/238 :)

Issue #710, opened 2023-03-02
"ipconfigX" recreates VM


saturn4er:
ipconfig0,ipconfig1... etc... is cloud-init parameter and can be changed without vm recreation, so why ForceNew is set to true for them?

PR #713, opened 2023-03-04
Remove ForceNew from ip option

saturn4er:
As I mentioned in #710, this options change do not require vm recreation.


mleone87:
merging but:

- solving the force new is ok since conceptually is wrong 
- cloudinit config will not change anyway if I recall correctly

PR #725, opened 2023-03-21
Add new vm_state attribute to manage running/stopped VMs

jonglezb:
This is useful in two cases:

- when some VMs should be stopped (for security reasons or maintenance): this is now possible from Terraform

- when a VM has been stopped manually in Proxmox: Terraform will now start the VM when applying the configuration

This deprecates the "oncreate" attribute following the guidelines from https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#renaming-an-optional-attribute


mleone87:
hello @jonglezb 

please notice that setting the vm_state as "stopped" when the vm is running will stop the vm as intended.
but setting the state as "running" when the vm is stopped does not start the vm

jonglezb:
Strange, it should work, I did test this feature before submitting (this is my main use-case). Can you give more information about how you run this? Any debug log?

mleone87:
that's easy to spot, maybe your are not using agent=1 parameter?

The actual code order in update function checks for the ip before starting the vm and goes on for 20min at least. then it arrives to the code that powers on the vm

I'll make a fix later today

PR #728, opened 2023-03-22
fixes #712 : destroy container fails "500 CT not running"

### LINKED ISSUE #712: Destroy container fails with "500 CT not running" error
I am new to this provider, and relatively inexperience with Terraform to be honest, but when trying to destroy a container that's not running (e.g., vmid 501) I get `500 CT 501 not running`.  I can see in my packet capture that it has 3 attempts to stop the container before giving up; that is 3 x `POST /api2/json/nodes/pve/lxc/501/status/stop`. If I start the container and try again all goes well. I've not tried any VMs yet, just CTs. 

I am using version `2.9.13` of the provider and rolling back to release `2.9.11` seems to fix this issues.

btorresgil:
Fixes #712

### Previous behavior

Destroy attempts to stop VM even if it's currently stopped. API return HTTP status 500 (see below) which is returned to provider causing it to error out.

```
HTTP/1.1 500 CT 110 not running
Connection: close
Content-Length: 13
Cache-Control: max-age=0
Content-Type: application/json;charset=UTF-8
Date: Wed, 22 Mar 2023 21:55:11 GMT
Expires: Wed, 22 Mar 2023 21:55:11 GMT
Pragma: no-cache
Server: pve-api-daemon/3.0

{"data":null}
```

### New behavior

Destroy checks if VM is `running` or `stopped` and only attempts to stop the VM if status is currently `running`.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -1702,24 +1702,30 @@ func resourceVmQemuDelete(d *schema.ResourceData, meta interface{}) error {
 	client := pconf.Client
 	vmId, _ := strconv.Atoi(path.Base(d.Id()))
 	vmr := pxapi.NewVmRef(vmId)
-	_, err := client.StopVm(vmr)
+	vmState, err := client.GetVmState(vmr)
 	if err != nil {
 		return err
 	}
-
-	// Wait until vm is stopped. Otherwise, deletion will fail.
-	// ugly way to wait 5 minutes(300s)
-	waited := 0
-	for waited < 300 {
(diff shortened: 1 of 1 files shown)


btorresgil:
Note: much easier to read diff with 'Hide whitespace' enabled. Very few lines actually changed, but indentation changed on several lines to accommodate new behavior.

mleone87:
nice spot, thanks

coaxial:
Thanks. Any idea when this will be released to the Terraform registry?

maxkratz:
Any schedule to release a new version of the provider to include this fix?

Issue #685, opened 2023-02-08
Network install (PXE) reboot indefinitely 


GannonTdW:
Hello,

Step to reproduce:
- create a main.tf with a PXE resource as described in [the documentation](https://registry.terraform.io/providers/Telmate/proxmox/latest/docs/resources/vm_qemu#provision-through-pxe-network-boot):
```terraform
resource "proxmox_vm_qemu" "pxe-minimal-example" {
    name                      = "pxe-minimal-example"
    agent                     = 0
    boot                      = "order=net0;scsi0"
    pxe                       = true
    target_node               = "test"
    network {
        bridge    = "vmbr0"
        firewall  = false
        link_down = false
        model     = "e1000"
    }
}
```
- `terraform apply`
- the resource is created, boots on the network, builds and reboots ... but like `boot = "order=net0;scsi0"`, the VM still boots on the network and restarts a build cycle and reboots etc

I have create a VM "by hand", with the boot order `scsi0,net0` -> No issue. The boot on the  empty scsi0, then on the network, builds, reboot on scsi0.

I tried to use `boot = "order=scsi0;net0"`, but I get the following error

>  Error: no network boot option matched in 'boot' config

Versions:
- Proxmox: 7.3-4
- terraform: v1.3.7 (linux)
- terraform-provider-proxmox: 2.9.11
- VM OS: Fedora CoreOS 37.20221225.3.0

tnyeanderson:
This should be fixed in 2.9.13 by #668 

Could you upgrade and try again?

GannonTdW:
I'm not sure if this is related to your changes but now I have this error

> proxmox_vm_qemu.pxe-minimal-example: Creating...
╷
│ Error: cloud-init parameters only supported on clones or updates
│
│   with proxmox_vm_qemu.pxe-minimal-example,
│   on main.tf line 22, in resource "proxmox_vm_qemu" "pxe-minimal-example":
│   22: resource "proxmox_vm_qemu" "pxe-minimal-example" {
│
╵
The main.tf part:
```
resource "proxmox_vm_qemu" "pxe-minimal-example" {
    name                      = "pxe-minimal-example"
    agent                     = 0
    boot                      = "order=net0;scsi0"
    pxe                       = true
    target_node = var.target_node
    network {
        bridge    = "vmbr0"
        firewall  = false
        link_down = false
        model     = "e1000"
    }
}
```

tnyeanderson:
[...]

PR #729, opened 2023-03-22
change PXE boot order and clean

GannonTdW:
Once the OS is installed, you have to boot on the disk.
With the opposite, it will always boot on the network

Cf: Issue https://github.com/Telmate/terraform-provider-proxmox/issues/685

> Changing some of the documentation is good.
> Out changing the documentation and the code is better

PS: I don't understand why the part between `<del>` and `</del>` is there. So I delete.


GannonTdW:
I have never learned go and I am not a dev.

Moreover, with this modification, if the network where the PXE server is located is on another network interface (e.g. net1), it will not work.
you would need something like this:
> scsi[0-9];net[0-9]

Issue #341, opened 2021-05-04
Feature request: support EFI disks


jqueuniet:
To properly work, an UEFI Proxmox VM needs a special disk called the EFI disk.

https://pve.proxmox.com/pve-docs/chapter-qm.html#qm_bios_and_uefi

The disk sub-block should be modified to allow creation of this resource. It only asks for two parameters, storage and type. The size is fixed.

Something like:

```
  disk {
    type = "efi"
    format = "qcow2"
    storage = "local"
  }
```

pgporada:
The upstream api will eventually get this functionality then it can be added to this provider. https://github.com/Telmate/proxmox-api-go/pull/116

jqueuniet:
Good to know, thanks for the response and the linked issue.

KnKay:
Nice to see, I am not alone with this! ;)

AshDevFr:
Any update on this feature?

KnKay:
This is in. I added it in Mai or June I guess.,

brutesque:
Is there an example or documentation on how to use this in a terraform resource?

scottbot95:
@brutesque Given @KnKay's contribution to the API package, it is "now" (as of 42a98880) *somewhat* possible to define EFI disks. If you add a new disk entry and set the type to `efidisk`, it looks like it will result in the expected API call. Ex
```
disk {
    type = "efidisk"
    size  = "1M"
}
```

The reason I say it is only "somewhat" possible however is that unlike all the other possible types for a disk, `efidisk` has a special set of options. Specifically this means it is not currently possible to set the `efitype` or `pre-enrolled-keys` so hopefully the PVE defaults work for your use-case. Additionally the lack of documentation in this package has meant that despite this being possible for nearly a year and a half, it appears that nearly nobody knew.

---

Given it is only valid to have at most one efidisk and the configuration is really only tangentially related to other disk types, I propose we should split `efidisk` into a new block within the qemu resource. We can then have the terraform schema validation ensure there is at most one efidisk and would allow for the separate configuration options. The result would look something like this:
```
resource "proxmox_vm_qemu" "efi_vm" {
    /* ... */
   efidisk {
      efitype = "4m"
      storage = "local"
      size = "1M"
      pre_enrolled_keys = 1
   }
}
```

Sadly I don't really know Go at all, but if this languishes long enough I might be motivated to learn it just to make a PR for this.

riverar:
[...]

PR #732, opened 2023-03-27
Add initial efidisk support

riverar:
Fixes: #341 

The `size` parameter was ignored as it appears proxmox only supports two flash types--2m and 4m--and creates a 4M backing file regardless.

The `pre-enrolled-keys` flag was skipped due to the work needed to snake_case the key back/forth to meet terraform requirements. The default value (1) seems satisfactory for most scenarios. We can revisit if the need arises.

Changing `efidisk` properties will force a replacement in this initial implementation. 

While it's _technically possible_ to remove the efidisk and replace it, there appears to be some quirky proxmox behavior in this area. Specifically, proxmox will accept updates to the efidisk but drop those changes on the floor. Proxmox also exhibits strange behavior when removing an efidisk from a running hotplug-enabled machine. (It will attempt to hot unplug, fail, but commit the change.) Given most users will create the efidisk and not touch it again, I opted for the simpler approach here. We can revisit if the need arises.

<details>
  <summary>Completed tasks</summary>

- [x] Investigate `pre-enrolled-keys` / `pre_enrolled_keys` mismatch
   Outcome: Not going to support this on first pass.
- [x] Determine if there's a more idiomatic way to handle `efidisk` other than using the first item in the slice `ExpandDevicesList` returns
   Outcome: Will defer to code review; added check to handle negative efidisk case.
- [x] Test create, destroy, and clone scenarios
  Outcome: Basic scenario tests successful.
- [x] Update docs
  Outcome: Docs updated.

</details>

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -413,6 +413,29 @@ func resourceVmQemu() *schema.Resource {
 					},
 				},
 			},
+			"efidisk": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Resource{
+					Schema: map[string]*schema.Schema{
+						"storage": {
+							Type:     schema.TypeString,
+							Required: true,
+							ForceNew: true,
+						},
+						"efitype": {
+							Type:     schema.TypeString,
+							Optional: true,
+							Default:  "4m",
+							ValidateFunc: validation.StringInSlice([]string{
+								"2m",
+								"4m",
+							}, false),
+							ForceNew: true,
+						},
+					},
+				},
+			},
 			"disk": {
 				Type:          schema.TypeList,
 				Optional:      true,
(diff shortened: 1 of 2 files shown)


riverar:
@mleone87 Anything needed on my part to get this over the finish line?

proxmox/resource_vm_qemu.go
```diff
@@ -1269,6 +1298,7 @@ func resourceVmQemuUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 		"bios",
 		"boot",
 		"bootdisk",
+		"efidisk",
```
mleone87, line 1301: Hello @riverar I think that the update resource update part needs more code ;)

mleone87:
hello @riverar did you have the opportunity to check my comments on the code?

riverar:
Not yet, my apologies for the silence, will find some time to revisit asap. We can mark this as a draft until ready if you'd like.

riverar:
Currently blocked on https://github.com/Telmate/proxmox-api-go/issues/253

brotherdust:
@riverar , I just ran into this issue. Any idea when this PR will be merged? Would like to default to EFI-only VM's. Also, are you aware of any way to set the EFI settings, such as disabling secure boot?

Thanks!

riverar:
Yeah this is blocking me too. Next week is bad for me but I should have lots of free time afterwards to see what's left to work on.

riverar:
OK, can pick this back up now. Sorry for the delay; had a few life events occur back to back.

riverar:
Friendly ping @mleone87, anything else needed here for an initial implementation?

PR #748, opened 2023-04-18
feat: cloud init disk resource

lsjostro:
This new resource creates/manage cloud-init configdrive as ISO-files. It generates a cloud-init iso and uploads it to proxmox. This enables a wider/more advanced support for cloud-init then Proxmox currently supports.

To use the generated cloud-init iso, VMs just need to define a `disk` block with `type=ide`, `media=cdrom` and `volume=proxmox_cloud_init_disk.<my_ci_resource>.id` parameters. 

Need to add docs before it's ready to be merged. But a first review would be helpful.

example:
```
resource "proxmox_cloud_init_disk" "ci" {
  name      = var.name
  pve_node  = var.pve_node
  storage   = "cephfs"

  meta_data = <<EOF
instance-id: ${var.name}
local-hostname: ${var.name}
EOF

  user_data = <<EOF
#cloud-config
users:
  - name: foo
    sudo: ALL=(ALL) NOPASSWD:ALL
EOF
  
  network_config = <<EOF
...your network config...
EOF
}

resource "proxmox_vm_qemu" "foo" {
....
  disk {
    type    = "ide"
    media   = "cdrom"
    storage = "cephfs"
    volume  = proxmox_cloud_init_disk.ci.id
    size    = proxmox_cloud_init_disk.ci.size
  }
...
}

```


lsjostro:
added docs now, so I think it's ready for review!

proxmox/provider.go
```diff
@@ -156,10 +156,11 @@ func Provider() *schema.Provider {
...
 		ResourcesMap: map[string]*schema.Resource{
-			"proxmox_vm_qemu":  resourceVmQemu(),
-			"proxmox_lxc":      resourceLxc(),
-			"proxmox_lxc_disk": resourceLxcDisk(),
-			"proxmox_pool":     resourcePool(),
+			"proxmox_vm_qemu":    resourceVmQemu(),
+			"proxmox_lxc":        resourceLxc(),
+			"proxmox_lxc_disk":   resourceLxcDisk(),
+			"proxmox_pool":       resourcePool(),
+			"proxmox_cloud_init": resourceCloudInitDisk(),
```
mleone87, line 163: I would call the resource proxmox_cloud_init_disk

proxmox/resource_cloud_init.go
```diff
@@ -0,0 +1,174 @@
...
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+			"meta_data": {
+				Type:     schema.TypeString,
+				Optional: true,
+				ForceNew: true,
+			},
+			"network_config": {
```
mleone87, line 53: I would add "vendor_data" field https://canonical-cloud-init.readthedocs-hosted.com/en/latest/explanation/vendordata.html

It would be exactly the same as others but run only at first boot, proxmox support it too

proxmox/resource_cloud_init.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+
+	if !isoFound {
+		// ISO not found so we (re)create it
+		d.SetId("")
+		return nil
+	}
+
+	return nil
```
mleone87, line 160: I would return diag and not nil when possible

proxmox/resource_cloud_init.go
```diff
@@ -0,0 +1,174 @@
...
+	pconf := m.(*providerConfiguration)
+	client := pconf.Client
+
+	storage := strings.SplitN(d.Id(), ":", 2)[0]
+	isoURL := fmt.Sprintf("/nodes/%s/storage/%s/content/%s", d.Get("pve_node").(string), storage, d.Id())
+	err := client.Delete(isoURL)
+	if err != nil {
+		return diag.FromErr(err)
+	}
+	return nil
```
mleone87, line 173: I would return diag and not nil when possible

mleone87 (commented):
great contribution, I left some comments otherwise LGTM

proxmox/resource_cloud_init.go
lsjostro: as it's not a an error. We should just return nil right?

proxmox/resource_cloud_init.go
lsjostro: as above

proxmox/resource_cloud_init.go
mleone87: that should be ok, you're right

proxmox/resource_vm_qemu.go
```diff
@@ -1662,7 +1667,7 @@ func resourceVmQemuRead(ctx context.Context, d *schema.ResourceData, meta interf
 
 	// need to set cache because proxmox-api-go requires a value for cache but doesn't return a value for
 	// it when it is empty. thus if cache is "" then we should insert "none" instead for consistency
-	var rxCloudInitDrive = regexp.MustCompile(`^.*-cloudinit.*`)
+	var rxCloudInitDrive = regexp.MustCompile(`^vm-[0-9]+-cloudinit.raw$`)
```
mleone87, line 1670: I think that should be left as it is, .raw extension is only present with storage that supports it

lsjostro: ah! cheers! ran with debug a now I see it.

lsjostro:
before merge we prob need the fix in the go-api as well. https://github.com/Telmate/proxmox-api-go/pull/254

lsjostro:
I guess this can be merged then. Will not 100% work in all cases until we get the new reworked disk handling in the go API into this project.

paraglade:
Note: this update requires `Datastore.AllocateTemplate` permssions
from the Proxmox user manual:
https://pve.proxmox.com/pve-docs/chapter-pveum.html#_privileges:~:text=Datastore.AllocateTemplate%3A%20allocate/upload%20templates%20and%20ISO%20images

> Datastore.AllocateTemplate: allocate/upload templates and ISO images

PR #754, opened 2023-04-24
lxc: implement deletion of network interfaces

### LINKED ISSUE #753: Network interfaces (of LXCs) are not deleted
Disclaimer: Don't know if this is happening to VMs aswell.

When you remove a network interface in the code Terraform correctly shows the difference
```
 # module.lxcs.proxmox_lxc.tempCopy_100 will be updated in-place
  ~ resource "proxmox_lxc" "tempContainer" {
        id                   = "proxmox/lxc/100"
        # (25 unchanged attributes hidden)

      - network {
          - bridge   = "vmbr0" -> null
          - firewall = false -> null
          - hwaddr   = "B2:6A:15:63:1C:B1" -> null
          - ip       = "manual" -> null
          - mtu      = 0 -> null
          - name     = "eth1" -> null
          - rate     = 0 -> null
          - tag      = 0 -> null
          - type     = "veth" -> null
        }

        # (2 unchanged blocks hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```
but does not actually delete that interface.

The required API call is sth like this
```
curl -X PUT -k -b PVEAuthCookie= -H "CSRFPreventionToken:" -d delete=net20,net21 https://192.168.1.222:8006/api2/json/nodes/node1/qemu/177/config -H 'Authorization: PVEAPIToken=root@pam!api_token=9d6824c0-a1c9-453e-8524-9ce179120c3e'
```
(command is taken from [here](https://forum.proxmox.com/threads/how-to-remove-network-interface-using-proxmox-api.122595/))

There also is a [TODO in the code](https://github.com/Telmate/terraform-provider-proxmox/blob/eb30c2d21fbcbb50c4e4bd924f178904c1a24991/proxmox/resource_lxc.go#L670) for this.

GamerBene19:
Closes #753

Disclaimer: I did(/do) not know Go, so there probably is room for improvements. Please tell me how it can be improved  and I'll refactor.

### Approach

First I tried matching simply based on name (e.g. `eth0`) however I've found this not to be desirable since names don't have to be unique and a simple rename would lead to recreation of a interface.

I then switched to using the Proxmox `id` since it does not have these flaws. I'm not 100% happy with the fact that I now need to drop the ids before applying the config. Suggestions for a better approach are welcome.

I've tested it quite a bit but certainly more testing would be good. I also don't know the impact on already existing states (since id is now added and was not there before).

### DIFF
--- a/proxmox/resource_lxc.go
+++ b/proxmox/resource_lxc.go
@@ -232,6 +232,10 @@ func resourceLxc() *schema.Resource {
 				Optional: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"id": {
+							Type:     schema.TypeInt,
+							Computed: true,
+						},
 						"name": {
 							Type:     schema.TypeString,
 							Required: true,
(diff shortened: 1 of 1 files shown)


GamerBene19:
@mleone87 you seem to commit to this repo the most:
Is there anything else I need to do to get this PR reviewed (and ideally merged)?

PR #766, opened 2023-05-11
Added multiple target nodes instead of a single node.

PascalKu:
Imagine you have a cluster in which you migrate using High Availability Server. this is exactly the case I had. I have created a HA Group in Proxmox within which servers are automatically migrated in case of a failure or similar. 

With this code change it doesn't matter now. You can specify an array of possible nodes where the QEMU can be located. 

I also implemented my own change to your golang implementation (https://github.com/Telmate/proxmox-api-go/pull/262) to manage this with HA Groups.

### DIFF
--- a/proxmox/data_ha_group.go
+++ b/proxmox/data_ha_group.go
@@ -1,0 +1,67 @@
+package proxmox
+
+import (
+	"sort"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func DataHAGroup() *schema.Resource {
+	return &schema.Resource{
+		Read: dataHaGroupRead,
+		Schema: map[string]*schema.Schema{
+			"group_name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"nodes": {
+				Type:     schema.TypeList,
+				Computed: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
+			},
+			"type": {
+				Type:     schema.TypeString,
+				Computed: true,
+			},
+			"restricted": {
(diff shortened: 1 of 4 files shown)


aderumier:
Hi,

I'm currently working to fix it in the right ways proxmox side
(autofind nodename if /nodes/qemu/<vmid> is called)

https://lists.proxmox.com/pipermail/pve-devel/2023-May/057007.html

I'm trying to push for proxmox8 

you can follow it here:

https://bugzilla.proxmox.com/show_bug.cgi?id=4689

PascalKu:
It's been a long time since the last update in the feed. Anything happened in the past?

lug-gh:
@PascalKu how to use this feature? It was merged, but documentation hasn't been updated.
From the source code 
Based on the source code `Nodes:      strings.Split(itemMap["nodes"].(string), ","),` , I assume that I have to submit the hosts as a string, separated by a comma. `target_node = "pve1,pve2,pve3"

However, I then get the error "Error: 400 Parameter verification failed." when applying.
It is displayed in the plan without any problems.

PascalKu:
Hi @lug-gh - do you still need help with this?

lug-gh:
@PascalKu Yes, I still haven't figured it out.

moonbaseDelta:
Hi, is this working only for VMs? Why not with LXC?

PascalKu:
> Hi, is this working only for VMs? Why not with LXC?

Because I only use QEMU - and nobody else updated it for LXC aswell.

PascalKu:
> @PascalKu Yes, I still haven't figured it out.

I'll now try to help you in the opened issue.

PR #809, opened 2023-07-13
Adding the role SND.use required for lxc in Proxmox 8.0.3

hnrazevedo:
I'm practicing Terraform in my home lab with Proxmox 8.0.3 and in the creation of my first instance LXC I came across the error that it was necessary the SND.use function for such a resource, so I took the opportunity to contribute in the doc.


irish1986:
I think we should fix SDN.use to SDN.Use (capital U).

hnrazevedo:
> I think we should fix SDN.use to SDN.Use (capital U).

adjusted

PR #822, opened 2023-08-04
Add ISO Storage resource

h0lyalg0rithm:
Was trying to automate my setup and found this missing.
It add the ISO storage resource which accepts iso url and pushes it through the proxmox api


legendary-cookie:
This would be useful

PR #862, opened 2023-11-09
feat: added unlink of disks when removing disk blocks

FredrickB:
**Requires https://github.com/Telmate/proxmox-api-go/pull/279**

During an update, the provider will check for
disks previously attached to the VM which has
now been removed from the HCL. The provider
unlinks and detaches the disks if the disks are
no longer a part of the configuration.

### DIFF
--- a/proxmox/disk_detach_util.go
+++ b/proxmox/disk_detach_util.go
@@ -1,0 +1,57 @@
+package proxmox
+
+import (
+	"fmt"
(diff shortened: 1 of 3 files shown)


mleone87:
@FredrickB merge or close?

FredrickB:
This can be closed. Sorry that I haven't been able to follow up on it.

PR #878, opened 2023-12-13
Bugfix/require recent module

MaartendeKruijf:
Tiny fix to allow for building the provider again.


mleone87:
@MaartendeKruijf an update of the go modules and we are ready I think?

MaartendeKruijf:
@mleone87 I've updated the branch. On a side note I could also help maintain the provider if that is of some interest to you?

mleone87:
@MaartendeKruijf thanks! Any help is appreciated

MaartendeKruijf:
@mleone87 could you remove the email from the commit message?

PR #886, opened 2023-12-22
resource_vm_qemu: allow . (dot) in name of virtual machine

### LINKED ISSUE #885: Allow usage of dot (`.`) in virtual machine names
Since #858, it is no longer possible to use `.` in the name of a VM. I usually name my virtual machines to the FQDN of the virtual machine, like `vm1.example.com`, which is no longer possible with this module. I propose to allow `.` in the name (see the PR I've created, #886).

I haven't found anything in the Proxmox documentation that would hint that is not allowed to use `.` in the name of the VM. Please let me know if this is incorrect.

vilhelmprytz:
fixes #885

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -75,12 +75,12 @@ func resourceVmQemu() *schema.Resource {
 				Description: "The VM name",
 				ValidateFunc: func(val interface{}, key string) (warns []string, errs []error) {
 					v := val.(string)
-					matched, err := regexp.Match("[^a-zA-Z0-9-]", []byte(v))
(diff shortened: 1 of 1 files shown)


mleone87:
@vilhelmprytz thanks for you PR but it's not how regex works, in fact also the actual regex is totally broken

vilhelmprytz:
> @vilhelmprytz thanks for you PR but it's not how regex works, in fact also the actual regex is totally broken

@mleone87 Can you please clarify? I believe the regex itself is correct, but I don't know about the specific Go implementation.

mleone87:
@vilhelmprytz have you tested this regex?

vilhelmprytz:
@mleone87 Yes, I have!

Issue #904, opened 2024-01-20
[resourceVmQemuRead] "ipconfig: '': source data must be an array or slice, got map"


mleone87:
https://github.com/Telmate/terraform-provider-proxmox/blob/8f3f9dea09ecef84e3962c582b32a82c4f422214/proxmox/resource_vm_qemu.go#L1744

@Tinyblargon this part is not happy if using plain legacy ipconfig[0-16] configuration, but it should be easy fix. I'll take a look today

mleone87:
I wrongly called Tinyblargon there but the change in that part is from @spettinichi 
@spettinichi did you have the opportunity to test the code you submitted?

config.Ipfonfig is a map but the ipconfig schema is not compatibile with that

mleone87:
Feel free to submit a properly tested PR, thanks!

PR #905, opened 2024-01-21
Revert d02bb5d not properly tested

mleone87:


mleone87:
solves https://github.com/Telmate/terraform-provider-proxmox/issues/904

PR #916, opened 2024-01-28
Containerise terraform-provider-proxmox

clincha:
Hi all, I wanted to share something I've been working on.

This merge request adds the code for creating two Docker images and uploading them to Docker Hub. Both images contain the tagged version of the terraform-provider-proxmox. One additionally contains the Azure Resource Manager binary as an example of how to add dependencies. The images come with Terraform installed and run if no other entrypoint is given.

## What the code does

1. A maintainer pushes a new version tag to the Telmate/terraform-provider-proxmox repository (v1.0.0)
2. The GitHub action fires and logs in to Docker Hub before calling the make file for each container in the matrix on line 10
3. The container_build recipe starts 
    - podman pulls a container with Golang installed and builds proxmox-terraform-provider using the make "build" recipe
    - podman pulls a container with Terraform installed and copies the proxmox=terraform-provider into it
    - (optional) Additional dependencies are installed 
4. the container_push recipe uploads the image to Docker Hub.

## Why?

- Terraform and all your dependencies wrapped into one easy to use Docker image
- This image allows you to layer on your own Terraform code to create a point in time snapshot of your infrastructure. You could rebuild your infrastructure to any specific point in time (based on a commit/tag)
- Consistent test and build environments
- Trivial to change versions of Golang, Terraform, dependencies and the provider.
- Code can be prepared and run in one environment and then moved to another environment and perform consistently.

## How to use it

So if I wanted to use this image my workflow might look something like this:

1. Create a copy of the Dockerfile in the examples folder
2. Move it to the root of my Terraform code
3. Run this to build the container image
    ```bash
    podman build . --file container_example.Dockerfile --tag docker.io/clincha/my-infrastructure:1.0.0`
    ````
5. Run this to create a plan. Note that I'm passing in any variables starting with TF_VAR and also my Azure Resource Manager access key. 
    ```bash
    podman run --entrypoint sh --env="TF_VAR*" --env="ARM_ACCESS_KEY=${ARM_ACCESS_KEY}" docker.io/clincha/my-infrastructure:1.0.0 -c "terraform init && terraform plan"
    ````
---
Let me know what you think, we would probably need to create a Docker Hub account for Telmate and get the variable values changed over if you're wanting to merge it. I could also create a script to create the images for every existing version of the provider too. I'm using it in my infrastructure repository if you [want to see it in a live environment](https://github.com/clincha-org/clincha/blob/master/containers/Dockerfile).


clincha:
Abandoned @clincha

PR #929, opened 2024-02-03
[3.0.1-rc1] Feature: KMGT disk size

### LINKED ISSUE #912: [3.0.1-rc1] Consider revisiting disk size before release
The new disks format has changed the `size` parameter to be an `int` that represents "the size of the created disk in Gigabytes". This is a regression from the previous regex that supported `\d+[GMK]`. It also means it's not possible to create disks smaller than 1 gigabyte.

If supporting the previous regex expression is not desirable, perhaps an easier change would be to make size represent kilobytes to support folks needing disks smaller than 1GiB.

@Tinyblargon @mleone87

Tinyblargon:
Fixed #912.

Terraform half of: https://github.com/Telmate/proxmox-api-go/pull/305
Will update the go mode file once it is merged.


mleone87 (approved):

mleone87:
@Tinyblargon looks good!

Tinyblargon:
@mleone87 This is completed, also noticed the CI wont run on newly added commits.

riverar:
@Tinyblargon tibibytes should read tebibytes

PR #930, opened 2024-02-04
Fix parallel execution handling when returning early

### LINKED ISSUE #922: [3.0.1-rc1] Stuck at Still creating...
My config is below, I can't seem to get past "Still creating..." with agent set to both 1 or 0. I tried it with the root user and token user described in the install guide as well.  I don't see anything happen in the UI or any requests being made. I do see requests going out and getting 200 OK with: 
```
provider "proxmox" {
  pm_api_url = "${var.proxmoxURL}"
  pm_user = "root@pam"
  pm_password = "<password>"
  pm_log_enable = true
  pm_log_file   = "terraform-plugin-proxmox.log"
  pm_debug      = true
  pm_log_levels = {
    _default    = "debug"
    _capturelog = ""
  }
}
```

I'm using the following:
```
Proxmox 8.1.4
Terraform 1.7.1
Telmate/proxmox 3.0.1-rc1
```

```
  resource "proxmox_vm_qemu" "anima-cluster" {
  count       = var.clusterSize
  name        = "${var.prefix}${count.index + 1}"
  vmid        = var.vmidStartRange + count.index
  ipconfig0  = "ip=${var.networkID}.${count.index + var.networkIPStartRange}${var.networkCIDR},gw=${var.networkID}.1"
  nameserver = "${var.networkID}.1"
  searchdomain = "home.arpa"

  desc        = var.mdClusterDescription
  target_node = var.targetNode

  clone = var.cloudTemplate
  full_clone = true
  onboot = true

  cpu = "host"
  cores   = var.cores
  sockets = var.sockets
  memory  = var.memory
  agent = var.qemuAgent

  disks {
    scsi {
      scsi0 {
        disk {
          storage = var.storageDevice
          size = var.diskSize
          emulatessd = var.ssd
          discard = true
        }
      }
    }
  }

  network {
    bridge    = "vmbr0"
    model     = "virtio"
    firewall  = false
    link_down = false
  }

  os_type   = "cloud-init"
  ciuser = var.username
  cipassword = var.password
  sshkeys = var.sshKeys
}
```

```
proxmoxURL = "https://<domain>:8006/api2/json"
pmUser = "terraform-prov@pve!anima-cluster"
pmTokenSecret = "<token>"
cloudTemplate = "eterna:mantic"
prefix = "anima-"
clusterSize = 6
cores = 4 
sockets = 1
memory = 8192
qemuAgent = 0
targetNode = "pve"
networkID = "10.0.0"
networkCIDR = "/24"
networkIPStartRange = 9
vmidStartRange = 200
storageDevice = "eterna"
diskSize = 100
ssd = true
username = "<username>"
password = "<password>"
sshKeys = <<-EOT
ssh-ed25519 ...
EOT
mdClusterDescription = <<-EOF
# Home Lab Cluster
EOF
```

evan-a-a:
resourceVmQemuCreate and resourceVmQemuUpdate can both return early due to errors. When this happens, we must ensure that we "unlock" the slot we have claimed for parallel execution. This is safe to do in a defer action because even if we manually "unlock" prior to the function return, the unlock action checks to ensure that the "lock" is held prior to attempting to return a slot to the pool.

Fixes #922

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -805,7 +805,8 @@ func resourceVmQemuCreate(ctx context.Context, d *schema.ResourceData, meta inte
 
 	pconf := meta.(*providerConfiguration)
 	lock := pmParallelBegin(pconf)
-	// defer lock.unlock()
+	defer lock.unlock()
+
 	client := pconf.Client
 	vmName := d.Get("name").(string)
 	vga := d.Get("vga").(*schema.Set)
@@ -1054,6 +1055,7 @@ func resourceVmQemuUpdate(ctx context.Context, d *schema.ResourceData, meta inte
 	pconf := meta.(*providerConfiguration)
 	client := pconf.Client
 	lock := pmParallelBegin(pconf)
+	defer lock.unlock()
 
 	// create a logger for this function
 	logger, _ := CreateSubLogger("resource_vm_update")
@@ -1327,6 +1329,7 @@ func resourceVmQemuRead(ctx context.Context, d *schema.ResourceData, meta interf
 	pconf := meta.(*providerConfiguration)
 	lock := pmParallelBegin(pconf)
 	defer lock.unlock()
+
 	client := pconf.Client
 	// create a logger for this function
 	var diags diag.Diagnostics


evan-a-a:
The linked issue has a reproducer (appears to be the same issue). In general this is an issue when deploying more than `pm_parallel` resources and `pm_parallel` of them fail. So in my case, I am attempting to deploy 5 resources, and the first 4 enter failures, resulting in a deadlock. Note that this setup is more akin to a credit-based system, not locking a mutex (except for managing access to the accounting variable).

>AFAIK putting the defer inside the resourceVmQemuCreate/resourceVmQemuUpdate functions it's precisely the condition to break the limit criterion of pm_parallel since the mutex is unlocked even if the execution of the specific job has not yet finished

I'm not sure what you're getting at here. If we don't return the slot we've claimed back to the pool, then we'll eventually exhaust the available slots and end in deadlock.

mleone87 (approved):

mleone87:
@evan-a-a I can reproduce the issue with the current code.

If the number or resources created is more than `pm_parallel` and they fail, the code will hang indefinitely. This is easily reproducible for example creating vms with a `pool` that does not exist. 

If resources are `<= pm_parallel` execution of plan will fail with proper error. 
If resources are `> pm_parallel `the plan will start without doing anything(in my case with a bad `pool` value)

Your MR is in the right path like I was in the past(see the commented `defer`s) but is not enough.

This is easily spottable by testing the code with those changes, all the executions will hang as soon as the `resourceVmQemuRead` function is called because no slot is free for `resourceVmQemuRead` function to work.

I'll merge this MR but add explicit unlock before calling the read function.

PR #947, opened 2024-02-25
[RC2 fixes] fix cloudinit disk slot and pm_parallel oddities

mleone87:


proxmox/resource_vm_qemu.go
```diff
@@ -456,8 +456,8 @@ func resourceVmQemu() *schema.Resource {
 								Schema: map[string]*schema.Schema{
 									"ide0": schema_Ide("ide0"),
 									"ide1": schema_Ide("ide1"),
-									"ide2": schema_Ide("ide2"),
```
Tinyblargon, line 459: Why the change from ide3 to ide2?

docs/resources/vm_qemu.md
Tinyblargon, line 1: Somewhere in the docs it should say that `ide2` is reserve, because of the change withe the `cloud-init` disk.

proxmox/resource_vm_qemu.go
```diff
@@ -987,66 +987,59 @@ func resourceVmQemuCreate(ctx context.Context, d *schema.ResourceData, meta inte
...
-		// }
 	} else {
-		log.Print("[DEBUG][QemuVmCreate] vm_state != running, not starting VM")
+		logger.Info().Str("vmid", d.Id()).Msgf("vm_state != running, not starting VM")
 	}
 
 	d.Set("reboot_required", rebootRequired)
-	log.Print("[DEBUG][QemuVmCreate] vm creation done!")
+	logger.Info().Str("vmid", d.Id()).Msgf("VM creation done!")
 	lock.unlock()
```
Tinyblargon, line 1032: There is a defer unlock on 802,803

proxmox/resource_vm_qemu.go
```diff
@@ -1587,6 +1566,7 @@ func resourceVmQemuDelete(ctx context.Context, d *schema.ResourceData, meta inte
 	}
 
 	_, err = client.DeleteVm(vmr)
+	lock.unlock()
```
Tinyblargon, line 1569: There is a defer unlock on 1320,1299

proxmox/resource_vm_qemu.go
Tinyblargon, line 1: there are some `lock.unlock()` in there that are unnecessary due to deferred unlocks. Don't know for sure what happens when you try to unlock an unlocked mutex.

Tinyblargon (commented):
Confused about changing the `cloud-init` disk to `ide2`, docs don't represent this change. Found some redundant unlocks which could cause issues.

proxmox/resource_vm_qemu.go
mleone87: am I right with this change since ide2 should be not used directly? Ora maybe it can be used but it conflicts with cloudinit is someone want to deploy a vm with a ISO file with terraform?

mleone87:
so @Tinyblargon regarding the unlock() thing, investigating the issue in https://github.com/Telmate/terraform-provider-proxmox/pull/930 I found that if we use defer and return the create/update functions into read, we are not freeing the slot because the unlock is AFTER read ends... but the reads do not complete because we have no free slots.

so i leave the defer for failure INTO the execution of the create/update and i explicit unlock before read if all goes well. reading the docs about `defer something()` I understood is acceptable to be redundant using `something()`

Tinyblargon:
@mleone87 i get what you mean with calling the read from inside the update and create. instead of managing all the nested locks wouldn't it be easier to create a `resourceVmQemuReadNoLock()` and we call that form inside the create and update function.




```go
func resourceVmQemuRead(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
  pconf := meta.(*providerConfiguration)
  lock := pmParallelBegin(pconf)
  defer lock.unlock()
  return resourceVmQemuReadNoLock(ctx, d, meta)
}

func resourceVmQemuReadNoLock(ctx context.Context, d *schema.ResourceData, meta interface{}) diag.Diagnostics {
  // everything that is in the current version of resourceVmQemuRead()
}
```

mleone87:
thinking about your proposed solution @Tinyblargon it will not work either since there are no free slot when entering the read function

I tested my code and it worked well in various scenarios

proxmox/resource_vm_qemu.go
Tinyblargon: Well `ide2` used to be reserved for the iso propertie, but that was removed in #937. Currently only `ide3` is reserved for `cloud-init` but we could move that to the new `disks` schema as well and let people chose on which device they'll mount the `cloud-init` disk. Fro now i would leave it as `ide3` as it's more intuitive that ide 0,1,2 are usable instead of 0,1,3.

mleone87:
I should be ready to merge this

Tinyblargon:
@mleone87 still a bit confused about the change of the `cloud-init` disk. https://github.com/Telmate/terraform-provider-proxmox/pull/947#discussion_r1523905925 everything else looks good,

mleone87:
@Tinyblargon at the end I think that the only part to keep is the pm_parallel fix, if not fixed yet!

Tinyblargon:
@mleone87 Yes, everything else has been superseded by other pull requests.

theOmeletDev:
I this error related ? `Warning: slot: ide2 size is ignored when type = cloudinit` But I have not set a size for ide2

PR #988, opened 2024-04-19
fixes

spettinichi:
@hestiahacker here is my old PR with just the bug fixes


spettinichi:
I'm not sure how to actually trigger these, but my go linter flagged them and yeah they couldn't never catch as is, so this should just fix error checking, the only way it would break things is if something was incorrectly passing or these error checks are invalid. I just wanted to make sure at least one other person looked it over for any obvious issues

hestiahacker:
Code changes look good to me.

It looks like a closing parenthesis was inadvertently removed here: https://github.com/Telmate/terraform-provider-proxmox/pull/988/files#diff-4c5567d861ab20b6f7b3543602fdea0c567cbb189952048608b0bc170d11fa9fL124

spettinichi:
Good catch! Fixed that, and a few things the bot flagged. It flagged a ton of "dont use underscores in variable names" and its so much to adjust I wasn't sure I wanted to cover all of that in here, so that's all that is left that the bot flagged.

Tinyblargon:
@spettinichi Thank you for adding error handling. the error `dont use underscores in variable names` can be ignored. we use undersores in variable names as we aren't using Domain Driven Design.

PR #994, opened 2024-04-24
Feature: Add `protection` property to the `resource_vm_qemu`

xserrat:
Expose the `protection` property for the `resource_vm_qemu` already added in the API https://github.com/Telmate/proxmox-api-go/pull/320.


xserrat:
> @xserrat Thank you for adding this feature, I've tested it and it works as intended. However it's not possible to use `terraform destroy` when `protection` is `true`, I'll fix this upstream [Telmate/proxmox-api-go#330](https://github.com/Telmate/proxmox-api-go/issues/330).

Great! I've just realised 😄  Thank you too for this great provider 👏🏽

Issue #1078, opened 2024-08-25
Feature: Add management of resource pools


Sonlis:
## Description

Add the feature to manage cluster's resource pools. This feature would allow to read, create, modify and delete resource pools.

## Relevant PVE API path
`/api2/json/pools`

## How would the resource look like

```hcl
resource "proxmox_pool" "example_pool" {
  pool_id = "example" # required
  comment = "A pool to enable easier management of permissions"
}
```

If this feature is accepted, I can work on it and open a PR.

Sonlis:
Right after creating this PR, I checked the source code and realised the feature already exist 🤡 . That would teach me to completely trust the documentation and not check the source code before opening an issue. Feel free to close to issue 😄

Tinyblargon:
@Sonlis the pool feature was built, but never communicated it was built/released. It also uses a lot of deprecated code, it should use the code from https://github.com/Telmate/proxmox-api-go/blob/b6deb236b502b5eb20fb8e192855aaf3f22a42d0/proxmox/config_pool.go as this has proper error handling. Also, we shouldn't let users update the pool members with the pool resource. This is already done by the qemu and lxc resource and would just create state issues.
The schema you suggested is pretty much how i would do it, maybe change `pool_id` to `name` or `pool_name`, but that's just semantics.

PR #1086, opened 2024-08-29
Refactor/use updated get method

Sonlis:
Closes https://github.com/Telmate/terraform-provider-proxmox/issues/1078

Create, delete and update were already using the non-deprecated code, only read was. Modify it so it uses the updated code.

Also document the `pool` resource.


Tinyblargon:
@Sonlis Tested this, and everything functions as intended. Thanks for your contribution.

PR #1093, opened 2024-09-02
Enable starting an lxc container which was just created

Sonlis:
Setting the boolean value would have no effect when creating the lxc.

This is because the check to start it was in an if block, which was executed only if the container was updated, not created.

Move the start of the container out of the if block, so it is checked everytime.


Tinyblargon:
@Sonlis Thank you for your contribution.

Issue #768, opened 2023-05-13
Feature request: Clone by VM ID rather than name


chrisbenincasa:
It is possible to have multiple VMs of the same name, but IDs are unique.

The code chooses the first one: https://github.com/Telmate/terraform-provider-proxmox/blob/bd28855b7c34d9d6913f4ac22c1933610a085cc2/proxmox/resource_vm_qemu.go#L1028

Of course it's not ideal to have multiple VMs with the same name, but it's likely. Especially when using the packer plugin for generating templates, due to the restrictions put in place on the name: https://github.com/hashicorp/packer-plugin-proxmox/blob/main/builder/proxmox/common/config.go#L281

pipelinedave:
I agree. To me it was even kind of confusing at first that the name seems to be the unique identifier, since the VM ID would to me seem like a much more obvious and logical choice.

chrisbenincasa:
Issue still persists in latest version because this feature is unimplemented.

pipelinedave:
Yeah this keeps us from committing to using this as well

PR #1146, opened 2024-11-07
feature: possibility to clone vm based on id

machiiine:
Here is a proposition to be able to clone a VM based on id too as requested [here](https://github.com/Telmate/terraform-provider-proxmox/issues/768)


Tinyblargon:
@machiiine Thank you for your contribution.
I've tested it and everything works as expected.

PR #1157, opened 2024-11-18
Added x-vga option for using primary-gpu configuration.

amirsalehag:
its needed for being able to use gpu for creating vm's with a gpu.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -415,6 +415,11 @@ func resourceVmQemu() *schema.Resource {
 							Type:     schema.TypeInt,
 							Optional: true,
 						},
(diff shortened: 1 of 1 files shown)


Tinyblargon:
@amirsalehag Thank you but currently I'm re-implementing pci so it can use all setting available in the PVE GUI. So it will have support for setting the GPU. My plan is to get it done by the end of the week.

Tinyblargon:
My plan is to create a release this Sunday, pci is 95% finished. One pull request needs to be rebased, and there is one bug fix i have to commit.

amirsalehag:
@Tinyblargon 
so using the gpu should be no problem after that right? if so, i can stop its process and work on something else by then.

Issue #1172, opened 2024-11-25
Error: 500 CT 241 already running, `v3.0.1-rc5`


AobaIwaki123:
# Abstract

I faced an Error `Error: 500 CT 241 already running` when I create Proxmox LXC with version `v3.0.1-rc5`
After `terraform apply`, I got this error but Proxmox LXC is working properly.
So, I guess health check system is something wrong.
And also, I can not do `terraform destroy` on this LXC for some reason.
It's really inconvenient. I'm really grad if it's fixed 

## Checked

- [x] `v3.0.1-rc4` does not have this issue
- [x] If `start=false`, I can not see this issue. I can do `terraform apply` and `terraform destroy`
- [x] `v3.0.1-rc6` has same issue

# Full Error Log

```sh
module.proxmox_lxc.proxmox_lxc.basic: Creating...
module.proxmox_lxc.proxmox_lxc.basic: Still creating... [10s elapsed]

│ Error: 500 CT 241 already running
│ 
│   with module.proxmox_lxc.proxmox_lxc.basic,
│   on ../../modules/proxmox_lxc/main.tf line 1, in resource "proxmox_lxc" "basic":
│    1: resource "proxmox_lxc" "basic" {
```

# Environment

Here's my environment
If you need more info, please tell me.

- Ubuntu: 24.10
- Terraform: v1.9.8
- tfenv: 3.0.0
- Proxmox: 8.3.0
- proxmox provider: 3.0.1-rc5

# My Proxmox LXC Module

```hcl
resource "proxmox_lxc" "basic" {
  features {
      nesting = true
  }
  
  vmid = var.vmid
  target_node  = var.target_node
  hostname     = var.hostname
  ssh_public_keys = var.public_key

  ostemplate   = "local:vztmpl/${var.template}.tar.zst"

  cores  = var.cores
  memory = var.memory

  unprivileged = true
  start       = true

  // Terraform will crash without rootfs defined
  rootfs {
    storage = var.storage
    size    = "${var.disk_size}G"
  }

    network {
        name = "eth0"
        bridge = var.bridge
        ip = "${var.ip_address}/24"
        gw = var.gateway
    }
}
```

AobaIwaki123:
I could not see this issue on `3.0.1-rc6`. So, I close this issue

AobaIwaki123:
Sorry I did not update provider version from `rc4` and I found same issue on `rc6`.
So, I reopen this issue

[...]

PR #1227, opened 2025-01-18
Start LXC on create/update without error

BI0M:
Issue: https://github.com/Telmate/terraform-provider-proxmox/issues/1172

The `start` value is already submitted to the Proxmox API in the `CreateLxcContainer` function in client.go, meaning the changes made in #1093 would result in following error: Error: 500 CT 100 already running. 

Tested changes with with 3.0.1-rc6


bdruth:
Nice, thanks for pushing this! Looking forward to the merge so I can update the provider again! 😄 

👍

PR #1263, opened 2025-03-11
Feat tpmstate efi secureboot

galthaus:
This MR adds features in two commits.

1. pre_enrolled_keys to the efidisk stanza.  This gets translated to the pre-enrolled-keys flag to turn on secure boot.
2. Add tpmstate stanza to configure the tpm in the VM.  This takes a storage location and a version.

These were tested against v3.0.1-rc6.  The master tree has additional issues.

 The code difference between the two points on the tree is the pxapi rename to pveSDK.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -394,6 +394,12 @@ func resourceVmQemu() *schema.Resource {
 							Required: true,
 							ForceNew: true,
 						},
+						"pre_enrolled_keys": {
+							Type:     schema.TypeBool,
+							Optional: true,
+							Default:  false,
+							ForceNew: true,
+						},
 						"efitype": {
 							Type:     schema.TypeString,
 							Optional: true,
@@ -1506,6 +1512,10 @@ func ExpandDevicesList(deviceList []interface{}) (pveSDK.QemuDevices, error) {
 		// this is a map of string->interface, loop over it and move it into
 		// the qemuDevices struct
 		for configuration, value := range thisDeviceMap {
+			// XXX: Not sure where to put these
+			if configuration == "pre_enrolled_keys" {
+				configuration = "pre-enrolled-keys"
+			}
 			thisExpandedDevice[configuration] = value
 		}
 
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -34,6 +34,7 @@ import (
 	"github.com/Telmate/terraform-provider-proxmox/v2/proxmox/Internal/resource/guest/qemu/network"
 	"github.com/Telmate/terraform-provider-proxmox/v2/proxmox/Internal/resource/guest/qemu/pci"
 	"github.com/Telmate/terraform-provider-proxmox/v2/proxmox/Internal/resource/guest/qemu/serial"
+	"github.com/Telmate/terraform-provider-proxmox/v2/proxmox/Internal/resource/guest/qemu/tpm"
 	"github.com/Telmate/terraform-provider-proxmox/v2/proxmox/Internal/resource/guest/qemu/usb"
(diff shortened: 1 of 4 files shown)


sygibson:
@galthaus - thank you for putting this together - we really need this in our use case - looking forward to this getting pulled in !!

proxmox/resource_vm_qemu.go
```diff
@@ -384,6 +384,29 @@ func resourceVmQemu() *schema.Resource {
 			pci.RootLegacyPCI: pci.SchemaLegacyPCI(),
 			pci.RootPCI:       pci.SchemaPCI(),
 			pci.RootPCIs:      pci.SchemaPCIs(),
+			"tpmstate": {
```
Tinyblargon, line 387: Could you create a `tpm` folder in https://github.com/Telmate/terraform-provider-proxmox/tree/master/proxmox/Internal/resource/guest/qemu and move this code (L387-L409) to that package. The serial package explains the structure https://github.com/Telmate/terraform-provider-proxmox/tree/master/proxmox/Internal/resource/guest/qemu/serial

proxmox/resource_vm_qemu.go
```diff
@@ -1482,8 +1514,26 @@ func ReadSmbiosArgs(smbios string) []interface{} {
 	return append(smbiosArgs, smbiosMap)
 }
 
+func ExpandTPMList(tmpList []interface{}) (*pveSDK.TpmState, error) {
```
Tinyblargon, line 1517: Could you create a `tpm` folder in https://github.com/Telmate/terraform-provider-proxmox/tree/master/proxmox/Internal/resource/guest/qemu and move this code (L1517-L1534) to that package. The serial package explains the structure https://github.com/Telmate/terraform-provider-proxmox/tree/master/proxmox/Internal/resource/guest/qemu/serial

Tinyblargon (changes requested):
Thank you so much for your contribution.

Sorry to ask you to move the code to a different place, but this file in particular kept becoming difficult to maintain due to the amount of code in it.

Further there are 2 minor schema changes.

proxmox/resource_vm_qemu.go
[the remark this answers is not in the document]
galthaus: Will do

proxmox/resource_vm_qemu.go
```diff
@@ -384,6 +384,29 @@ func resourceVmQemu() *schema.Resource {
 			pci.RootLegacyPCI: pci.SchemaLegacyPCI(),
 			pci.RootPCI:       pci.SchemaPCI(),
 			pci.RootPCIs:      pci.SchemaPCIs(),
+			"tpmstate": {
+				Type:     schema.TypeList,
```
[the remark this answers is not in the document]
galthaus: Will do and removed length check.

proxmox/resource_vm_qemu.go
galthaus: Will do.  cleaner.

proxmox/resource_vm_qemu.go
galthaus: done

galthaus:
Updated and tested based upon 301-rc6.

Testing against master breaks with all sorts of errors.

For example: the ParseMAC calls will fail because ParseMAC only parses lowercase mac address.  All calls should be called with strings.ToLower.  This causes net processing to fail on update because proxmox returns up-cased macs.

For example:  the SDK library crashes at line 650 in config_qemu.go because the vmr structure is filled in after it is used.  Flopping these stanzas resolve this issue.

There is another plugin crash that I hadn't tracked down.  I'm quite certain they aren't related to this MR/PR.

It seems like master is not in a usable state for qemu resources.

Tinyblargon:
@galthaus did some testing it seems all right. Thank you for this contribution.

Issue #1236, opened 2025-01-31
regression: No way to set 'import-from' parameter since 3.0.1-rc2


TomyLobo:
#606 introduced an import_from parameter to disks, which added the `import-form` parameter to, for instance, the --scsi[n] option.
This was removed in https://github.com/Telmate/terraform-provider-proxmox/commit/5b89c2ea9c23b7dfbbcb047157f6679b1631d656#diff-104365919f693375882979581d9f36f5266991eef361948e56c220931fe886ddL600 and never re-added.

Is it possible to add this back?
It's really useful in order to import official disk images.

Tinyblargon:
@TomyLobo agree that we should at it back, but not sure yet about it's exact behavior and implementation. Pretty sure we need to add support for this in the upstream SDK.

Tinyblargon:
@TomyLobo that's of the old implementation. The new structure is this one https://github.com/Telmate/proxmox-api-go/blob/0f3daee4933465179889336911239ea249e5a7fb/proxmox/config_qemu_disk.go#L1027

TomyLobo:
any news on this?

cdkl:
I'm also interested in this, as I can't seem to find another way to provision a disk from an official qcow2 image, and what hints I've found all point back to `import_from`.

noahlehmann:
> [@TomyLobo](https://github.com/TomyLobo) agree that we should at it back, but not sure yet about it's exact behavior and implementation. Pretty sure we need to add support for this in the upstream SDK.

@Tinyblargon  - seems like it. I kinda need this feature, so I created a PR in the upstream project.
https://github.com/Telmate/proxmox-api-go/pull/412

Let's see if I missed anything. 
I'd be happy to add the feature here as well if needed and wanted - as soon as the upstream is updated.

Tinyblargon:
@noahlehmann #1306 has been merged, my restructuring of the qemu `disk` and `disks` is done.

noahlehmann:
@Tinyblargon thanks, hoping to get to it next week!

PR #1330, opened 2025-05-26
feature/import-disk: add import_from to disk options

noahlehmann:
This PR will resolve https://github.com/Telmate/terraform-provider-proxmox/issues/1236.

The option ìmport_from` was added in https://github.com/Telmate/terraform-provider-proxmox/pull/606 but never made the cut to the version `3.x.x` updates made to the provider. 

Simply adding the option to the SDK (which added that feature with https://github.com/Telmate/proxmox-api-go/pull/412) will create the disks.

However the size of the disk will not be changed, even if a different size than the imported disk is desired. 
A second `plan` will show that the sizes need to be changed (and `applied`) again.

There are still open points to be done here:
- [ ] ignore `import_from` for present disks, as proxmox doesn't save this option and won't return it either. It is only used for disk creation.
- [ ] let the TF provider resize the disk after creation so the state fits the planned resources.

### DIFF
--- a/proxmox/Internal/resource/guest/qemu/disk/schema.go
+++ b/proxmox/Internal/resource/guest/qemu/disk/schema.go
@@ -30,6 +30,7 @@ const (
 	schemaFormat            string = "format"
 	schemaID                string = "id"
 	schemaIDE               string = "ide"
+	schemaImportFrom        string = "import_from"
 	schemaIOPSrBurst        string = "iops_r_burst"
 	schemaIOPSrBurstLength  string = "iops_r_burst_length"
 	schemaIOPSrConcurrent   string = "iops_r_concurrent"
@@ -97,6 +98,7 @@ func SchemaDisk() *schema.Schema {
 				schemaEmulateSSD:        {Type: schema.TypeBool, Optional: true},
 				schemaFormat:            subSchemaDiskFormat(schema.Schema{}),
 				schemaID:                subSchemaDiskId(),
+				schemaImportFrom:        {Type: schema.TypeString, Optional: true},
 				schemaIOPSrBurst:        subSchemaDiskBandwidthIopsBurst(),
 				schemaIOPSrBurstLength:  subSchemaDiskBandwidthIopsBurstLength(),
 				schemaIOPSrConcurrent:   subSchemaDiskBandwidthIopsConcurrent(),
@@ -635,6 +637,7 @@ func subSchemaIde(slot string) *schema.Schema {
 							schemaLinkedDiskId:  subSchemaLinkedDiskId(),
 							schemaReplicate:     {Type: schema.TypeBool, Optional: true},
 							schemaSerial:        subSchemaDiskSerial(),
+							schemaImportFrom:    {Type: schema.TypeString, Optional: true},
 							schemaSize:          subSchemaDiskSize(schema.Schema{Required: true}),
 							schemaStorage:       subSchemaDiskStorage(schema.Schema{Required: true}),
 							schemaWorldWideName: subSchemaDiskWWN()})}},
@@ -718,6 +721,7 @@ func subSchemaSata(slot string) *schema.Schema {
 							schemaEmulateSSD:    {Type: schema.TypeBool, Optional: true},
 							schemaFormat:        subSchemaDiskFormat(schema.Schema{Default: "raw"}),
 							schemaID:            subSchemaDiskId(),
+							schemaImportFrom:    {Type: schema.TypeString, Optional: true},
 							schemaLinkedDiskId:  subSchemaLinkedDiskId(),
 							schemaReplicate:     {Type: schema.TypeBool, Optional: true},
 							schemaSerial:        subSchemaDiskSerial(),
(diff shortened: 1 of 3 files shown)


noahlehmann:
### TL;DR;

This took me way longer than expected, the nesting of `disk` but especially `disks` and the way TF core and provider SDK handle the state and the passed `ResourceData` almost drove me insane. 
Tested manually for now, resizing the disks is still a TODO.
Won't get to continue until end of next week. 

@Tinyblargon some of this is rather messy, let me know what you think and if you have any other approaches. 

### Current State

For now the `import_from` value is ignored from the API, the value is set as `WriteOnly = true`. This prevents it from being set to state and therefore not being persistet, as the TF Core/SDK compare the saved state with the values fetched from the Proxmox SDK, which always returns `null` for `import_from`, as the API won't return a value for that either (one-time value).

The implementation **ALWAYS** passes the `import_from` value to the SDK and therefore to the Proxmox API, but the API ignores that for already created disks. My conclusion to this is we can safely continue to do so.

### Regarding the `WriteOnly` attribute

See: https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/write-only-arguments

Unfortunately we need to use some rather experimental features of the v2 SDK (reading the raw config) to actually use the `import_from` when set to `WriteOnly = true`.
The nested structure and the possibility to have redundant `disk` or `disks` fields (continuing in the nesting) makes it hard to have an elegant solution (though honestly it might be my lack of the Golang knowledge).

See the `proxmox/Internal/resource/qemu/disk/sdk` changes here: https://github.com/Telmate/terraform-provider-proxmox/pull/1330/commits/9ba620ebe6da45601b0137641e18ed35167f541c

As far as I have tested this it works reliably - as long as the assumption that the nested lists always only have one attribute (e.g. `disks.0.ide.0.ide0.0.disk.0.import_from`) is true.

### Todo

The disks will still only have the size of the imported image, a second plan and apply will fix that, but this should be addressed.

Tinyblargon:
Thank you for your work on this feature.

A few points of feedback.

## WriteOnly

We get around using `WriteOnly` for the password in CloudInit with this trick:
https://github.com/Telmate/terraform-provider-proxmox/blob/58d4429334bd3c6deae46e3ccf52011097222090/proxmox/Internal/resource/guest/qemu/cloudinit/terraform.go#L15 Your `WriteOnly` implementation is clearer and would be preferable.


## Setting ImportFrom

Why isn't the logic for setting the `import_from` handled in the following places instead of `mapWriteOnlyImportFromDisksToType`?
As I understand it, you loop through the disks and when `import_from` is set all the settings of the disk will be cleared and `import_from` is added back?

https://github.com/Telmate/terraform-provider-proxmox/blob/58d4429334bd3c6deae46e3ccf52011097222090/proxmox/Internal/resource/guest/qemu/disk/sdk_disk.go#L98-L118
https://github.com/Telmate/terraform-provider-proxmox/blob/58d4429334bd3c6deae46e3ccf52011097222090/proxmox/Internal/resource/guest/qemu/disk/sdk_disks.go#L77-L98

## Sugestion

Maybe with getting the previous value of the `disk` and `disks` we can check if the user changed `import_from`?
This could allow for importing a disk over an existing one when the source changes.

```go
oldVal, newVal := d.GetChange("my_setting")
```

## Upstream

> The disks will still only have the size of the imported image, a second plan and apply will fix that, but this should be addressed.

This feels like an issue that should be resolved in the SDK/library itself, not in Terraform.
Ideally, in the future, providing `ImportFrom` to the upstream SDK should trigger a full re-import of the disk.

noahlehmann:
@Tinyblargon thanks for taking the time to assist! Greatly appreciated. 
Before I continue, here are some thoughts on your replies.

---
> Why isn't the logic for setting the `import_from` handled in the following places instead of `mapWriteOnlyImportFromDisksToType`?

Mostly because I didn't want to change the functions parameters passed in `sdk_disks.go` and `sdk_disk.go`, as I would need to pass either the `ResourceData` as whole or the `ResourceData.GetRawConfigAt('RootDisk[s]')`.

E.g. for IDE disks (top level)

https://github.com/Telmate/terraform-provider-proxmox/blob/58d4429334bd3c6deae46e3ccf52011097222090/proxmox/Internal/resource/guest/qemu/disk/sdk_disk.go#L54

https://github.com/Telmate/terraform-provider-proxmox/blob/58d4429334bd3c6deae46e3ccf52011097222090/proxmox/Internal/resource/guest/qemu/disk/sdk_disks.go#L49

\+ the lower level functions and all these for the 3 other types of disks as well.

If you think passing the raw config could be helpful in the future, i could refactor. 

---
> As I understand it, you loop through the disks and when `import_from` is set all the settings of the disk will be cleared and `import_from` is added back?

The docs on `WriteOnly` unfortunately state that a value set as such needs to be read from the raw practitioners config, so thats why i loop the settings. I only set the `import_from` value though, all other settings are untouched.

---
> Maybe with getting the previous value of the `disk` and `disks` we can check if the user changed `import_from`?

If we keep `import_from` as `WriteOnly` we can't set the state for the value, we would need to handle this differently then, what's your thought on that?

> This could allow for importing a disk over an existing one when the source changes.

I do like the idea of that feature, it would allow users to use some sort of "immutable" cloud image and do upgrades via import, however i would probably set this as new feature when this is done, we do need to consider the `WriteOnly` point from above when preparing though. Maybe this should need some flag as well (like `overwrite_import_from`) as this would behave different as the PVE API. 

---
> This feels like an issue that should be resolved in the SDK/library itself, not in Terraform.
> Ideally, in the future, providing `ImportFrom` to the upstream SDK should trigger a full re-import of the disk.

It depends, if the Provider SDK should behave like the actual PVE API (which I as user would suppose) I would add this to the terraform logic. But that's personal preference.
We should decide on this though, because it either needs to be done in this PR or the upstream should be fixed first.

noahlehmann:
@Tinyblargon any updates on this? thoughts?

Daskan:
@Tinyblargon any updates on this? :)

PR #1332, opened 2025-06-01
fix(doc): use single quotes for PM_API_TOKEN_ID env var

rubenhoenle:
Ran into some problems with the docs because of the exclamation mark in bash. 

Works for me when using single quotes. So I thought I'll send a quick patch.


Tinyblargon:
Thanks for your contribution.

PR #1360, opened 2025-07-19
Docs: Update Terraform import docs & add missing docs entry for `pre_enrolled_keys`

0x464e:
It seems a docs entry was forgotten when #1263 was made. I added it.

Also, the docs section about Terraform import had incorrect example syntax.  
I also added an example which should be especially helpful for beginners.


Tinyblargon:
@0x464e Thank you for your contribution :)

PR #1371, opened 2025-07-30
feat: add use_primary_interface option to control IP detection behavior

sungjinfcc:
This PR adds a new optional flag use_primary_interface to toggle between retrieving the VM IP from only the primary network interface (default behavior) or from any available interface via QEMU agent.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -53,11 +53,11 @@ const (
 )
 
 const (
-	schemaAdditionalWait      = "additional_wait"
-	schemaAgentTimeout        = "agent_timeout"
-	schemaSkipIPv4            = "skip_ipv4"
-	schemaSkipIPv6            = "skip_ipv6"
-	schemaUsePrimaryInterface = "use_primary_interface"
+	schemaAdditionalWait = "additional_wait"
+	schemaAgentTimeout   = "agent_timeout"
+	schemaSkipIPv4       = "skip_ipv4"
+	schemaSkipIPv6       = "skip_ipv6"
+	schemaPrimaryNetwork = "primary_network"
 )
 
 func resourceVmQemu() *schema.Resource {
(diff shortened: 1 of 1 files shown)


Tinyblargon:
@sungjinfcc thank you for helping out with the project. There is however an issue with the feature you are trying to add:
It is adding a new setting that isn't very versatile, there are probably situations where a user wants to select the 3rd interface.

Would you be able and willing to make the following changes?

Rename the setting to `primary_network`?

Allow as input numbers and the id of the network interface.
Setting the id would use that interface as the primary.
Where as setting a number would pick the N th network inteface.

For example if you have the following interfaces: `net0` `net7` `net12`
And `primary_network` is `2` then it picks `net12`

sungjinfcc:
@Tinyblargon thank you for the feedback. I've enhanced the feature to be more versatile while preserving all original functionality.

**Enhanced `primary_network` setting:**
- `0` (default): Use primary interface
- `-1`: Try all available interfaces  
- `N` (positive): Use specific interface ID or Nth interface

**Examples:**
- `primary_network = 7` → Use net7 (specific ID)
- `primary_network = 2` → Use 2nd interface
- `primary_network = -1` → Try all interfaces

Tinyblargon:
@sungjinfcc Really like the edition of `-1` as I understand it correctly it will go through the interfaces until it finds one with a valid IP.

I don't get the following example, how do we differentiate between 7th interface and `net7` and between 2nd interface and `net2`?

>`primary_network = 7` → Use net7 (specific ID)
>`primary_network = 2` → Use 2nd interface

sungjinfcc:
@Tinyblargon Yes, `-1` will try all interfaces until it finds one with a valid IP

The logic prioritizes ID over position. For example, if VM has net0, net7, net12:
- `primary_network = 7` → Uses net7 (ID match)
- `primary_network = 2` → No net2 exists, uses 2nd interface (net7)
- `primary_network = 12` → Uses net12 (ID match)

Issue #1413, opened 2025-10-01
500 error creating VM Proxmox 9.0


TobiPeterG:
Hi there,

when I'm trying to create a VM using this provider with Proxmox 9, opentofu/terragrunt shows this error after some time:
```
module.proxmox_vm.proxmox_cloud_init_disk.ci: Creation complete after 6s [id=cephfs:iso/tf-ci-test.iso]
module.proxmox_vm.proxmox_vm_qemu.vm: Creating...
module.proxmox_vm.proxmox_vm_qemu.vm: Still creating... [10s elapsed]
module.proxmox_vm.proxmox_vm_qemu.vm: Still creating... [20s elapsed]
module.proxmox_vm.proxmox_vm_qemu.vm: Still creating... [30s elapsed]
module.proxmox_vm.proxmox_vm_qemu.vm: Still creating... [40s elapsed]
module.proxmox_vm.proxmox_vm_qemu.vm: Still creating... [50s elapsed]

Error: 500 Internal Server Error

  with module.proxmox_vm.proxmox_vm_qemu.vm,
  on .terraform/modules/proxmox_vm/main.tf line 26, in resource "proxmox_vm_qemu" "vm":
  26: resource "proxmox_vm_qemu" "vm" {

time=2025-10-01T09:11:34Z level=error msg=terraform invocation failed in /app/config/data-dir/repos/sci/deployment/virtual_machines/130/pcloud_vm_configs_infrastructure_251001_test/pcloud/vm_configs/infrastructure/251001_test/.terragrunt-cache/5TQt7dDahpgne4lqy8rtTCF4r8w/YDaC2-xxFNen9AN8ugH4TCU9LyQ prefix=[/app/config/data-dir/repos/sci/deployment/virtual_machines/130/pcloud_vm_configs_infrastructure_251001_test/pcloud/vm_configs/infrastructure/251001_test] 
time=2025-10-01T09:11:34Z level=error msg=1 error occurred:
	* [/app/config/data-dir/repos/sci/deployment/virtual_machines/130/pcloud_vm_configs_infrastructure_251001_test/pcloud/vm_configs/infrastructure/251001_test/.terragrunt-cache/5TQt7dDahpgne4lqy8rtTCF4r8w/YDaC2-xxFNen9AN8ugH4TCU9LyQ] exit status 1
```

On this node, I see this in the logs:
```
sudo tail -n +1 /var/log/pveproxy/access.log
[...]

PR #1415, opened 2025-10-02
Skip ipv6 and ipv4 if os-type="cloud-init" and no IP configured.

TobiPeterG:
This fixes https://github.com/Telmate/terraform-provider-proxmox/issues/1413

I will test the changes now and see if it works as expected.

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -1338,6 +1338,30 @@ func initConnInfo(ctx context.Context, d *schema.ResourceData, client *pveSDK.Cl
 		}
 		ciAgentEnabled = true
 	}
+	
+	skipIPv4 := d.Get(schemaSkipIPv4).(bool)
+	skipIPv6 := d.Get(schemaSkipIPv6).(bool)
+
+	// If os_type=cloud-init and ipconfig0 missing (or has no DHCP/static), skip ipv4 and ipv6
+	if strings.EqualFold(d.Get("os_type").(string), "cloud-init") && config != nil && config.CloudInit != nil {
+		// ipconfig0 existence
+		_, hasIP0 := config.CloudInit.NetworkInterfaces[pveSDK.QemuNetworkInterfaceID0]
+
+		// DHCP/static address probe
+		noAddr := false
+		if hasIP0 {
+			cicustom := config.CloudInit.Custom != nil && config.CloudInit.Custom.Network != nil
+			tmpConn := parseCloudInitInterface(config.CloudInit.NetworkInterfaces[pveSDK.QemuNetworkInterfaceID0], cicustom, false, false)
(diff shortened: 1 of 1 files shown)


TobiPeterG:
Tested it, this seems to fix the issue. Now the provider won't wait for the VM to come up with cloud-init.

However, I noticed that a destroy action might _not_ work and also cause a 500 error when the guest agent is not running in the VM yet. Maybe the error message during creation or destruction with agent = 1 should be extended to not only show the 500 error, but also show that the guest agent is enabled but might not be running yet on the VM.

TobiPeterG:
Hmm I installed the guest agent on the VM and tried to delete it but I got a 500 error and the VM is not destroyed. Will investigate

TobiPeterG:
Curious, on the server itself I don't see a 500 error.
It only made two calls:
::ffff:10.124.31.44 - OpenTofu@pve!OpenTofu [02/10/2025:11:44:28 +0000] "GET /api2/json/nodes/cx39/qemu/198/pending HTTP/1.1" 200 545
::ffff:10.124.31.44 - OpenTofu@pve!OpenTofu [02/10/2025:11:44:28 +0000] "GET /api2/json/nodes/cx39/qemu/198/status/current HTTP/1.1" 200 850

TobiPeterG:
The 500 error also occurs with your current master version, so I don't think I caused it. Can this be confirmed?

My main.tf that I used for my provider version and your master version:
```
provider "proxmox" {
  pm_api_url      = "OURAPIURL"
  pm_api_token_id = "OpenTofu@pve!OpenTofu"
  pm_api_token_secret = "OURTOKEN"
  pm_tls_insecure = true
}

  name         = "local-test"
  protection   = false
  target_nodes = ["OURNODENAME"]
  memory       = 4096
  scsihw       = "virtio-scsi-pci"
  clone        = "OURTEMPLATE"
  os_type      = "cloud-init"
  pool         = "OURPOOL"
  agent        = 1

  cpu {
    cores   = 4
    sockets = 1
    type    = "host"
  }

  network {
      id      = 0
      model   = "virtio"
      bridge  = "vmbr0"
      tag     = OURVLAN
  }

  disks {
    scsi {
      scsi0 {
        disk {
          size    = 40
          storage = "OURSTORAGE"
        }
      }
    }
    ide {
      ide2 {
        cdrom {
          iso = "OURSTORAGEPOOL:iso/OURISONAME.iso"
        }
      }
    }
  }
  lifecycle {
    ignore_changes = [
      agent_timeout,
      disks[0].ide[0].ide2[0].cdrom[0].iso,
      target_node,
      target_nodes,
      agent
    ]
  }
}
```

With this setup I also confirmed that VM creation works without problems with my provider and fails with yours.
VM deletion fails with both versions, but I'm very sure that's another issue.

PR #1428, opened 2025-10-16
chore(docs): clarify docs on VM.Monitor

ivan-penchev:
As a new Promox user and as Proxmox 9 is becoming the LTS, I am suggesting this clarification on the docs.

I am ashamed to say so... but... I did spend a couple of minutes trying to figure out why the command from the docs does not work when I copy pasted it. :/ 

And even though it was right there, on the next pagraph, then I was confused why when I copied the command for "Proxmox > 8:"... it said that the role does not exist.

**TLDR: Dummy engineer, wants to have two sections, so that you can just directly copy/paste from the docs.**

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -6,8 +6,8 @@ and [proxmox_lxc](resources/lxc.md).
 
 ## Creating the Proxmox user and role for terraform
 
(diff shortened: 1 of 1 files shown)


Tinyblargon:
@ivan-penchev Thank you for improving the docs.

PR #1461, opened 2025-11-24
fix: ID not cleared when guest deleted from gui

### LINKED ISSUE #1458: Error: vm '#' not found
I know this was discussed previously here: [https://github.com/Telmate/terraform-provider-proxmox/issues/546](https://github.com/Telmate/terraform-provider-proxmox/issues/546)

I would really like this reconsidered, we previously recreated our environment Promxox environment and wanted to recreate our LXC containers which are managed through Terraform and triggered through a github action

This caused errors:
```
Error: vm '1011' not found
Error: vm '1012' not found
Error: vm '1013' not found
Error: vm '1021' not found
Error: vm '1022' not found
...
```

This currently causes issues when applying or destroying

I understand this could be manually fixed by running `terraform state rm` for each missing LXC container, however this really makes Terraform less automated and requires a bit more manual effort

Typically with other providers, AWS for example, would recreate the resource that is missing rather than returning an error if it does not exist.

I am hoping this can be reconsidered as it would make this situation easier and less of a manual effort to fix in the future

Tinyblargon:
When the guest is removed outside of Terrafrom we get an error when trying to read the deleted guest.
We now clear the ID marking the guest as removed in Terraform.
This way the Guest can automatically be created on the next run.

Resolves #1458

### DIFF
--- a/proxmox/diag_warnigns.go
+++ b/proxmox/diag_warnigns.go
@@ -1,0 +1,14 @@
+package proxmox
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+)
+
+func resourceDriftDeletionDiagnostic(d *schema.ResourceData) diag.Diagnostic {
+	id := d.Id()
+	d.SetId("")
+	return diag.Diagnostic{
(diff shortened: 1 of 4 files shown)


Tinyblargon:
Only instances of `lock.unlock()` is in `defer lock.unlock()` the defers are directly after `lock := pmParallelBegin(pconf)`, the locking logic.

PR #1465, opened 2025-12-04
fix: Add missing ide3 slot validation in flat disk schema

aleskxyz:
## Description

Fixes missing `ide3` validation in the flat `disk` schema. The provider supports 4 IDE slots (ide0-ide3), but the validation only checked ide0-ide2, causing inconsistency with the nested `disks` schema.

## Changes

- Added `ide3` to slot validation switch case in `schema.go`
- Updated error message to include `ide3` in valid slots list

## Verification

All other code paths already support `ide3` (terraform_disk.go, sdk_disk.go, etc.). This change only adds the missing validation to match existing functionality.


Tinyblargon:
@aleskxyz Thank you for your contribution

PR #1400, opened 2025-09-14
Feat: error when `automatic_reboot` = false

Tinyblargon:
Change the behavior of `automatic_reboot` to give an error instead of a warning when we can't reboot, as the task we asked of Terraform could not be completed.

Removes the hard-coded reboot triggers in favor of letting the SDK handle it. For cloud-init setting we will trigger a reboot manually.

Closes #1393

PR #1477, opened 2026-01-17
fix: read hagroup value from Proxmox

nE0sIghT:
In the #1400 the call to `NewConfigQemuFromApi`  was replaced by call to  `NewActiveRawConfigQemuFromApi`, but `ReadVMHA` is never called from  `NewActiveRawConfigQemuFromApi`, so `hagroup` is never populated and the provider always tries to set the HA group to the configured value.

This change restores `ReadVMHA`  call before populating `hagroup` from `vmr`. I'm not sure whether it would be better to add `ReadVMHA` to `NewActiveRawConfigQemuFromApi` itself instead.


Tinyblargon:
@nE0sIghT Thank you for fixing this bug.


`ReadVMHA()` is not part of `NewActiveRawConfigQemuFromApi()` because they use different API endpoints.

PR #1478, opened 2026-01-17
fix: `rate` network value should be MBps

nE0sIghT:
After upgrading from v2 to v3 I noticed that provider reports strange diffs for the rate limit of my VMs, for example:

```
~ rate = 768000000 -> 768
```

Here `768` is the value I configured, and  the `768000000` on the left looks wrong. According to the docs, the `rate` value must be specified in MBps.


Tinyblargon:
@nE0sIghT The docs say it's Mbit, but it allows for floating point numbers so. `2.234` is a valid amount of Mbit. We convert it to Kbit so only the last part of the chain has to deal with the floating point numbers.

We do have way to many zeros for Kbit, so it might even be bits per second.

Tinyblargon:
The SDK counts it in Kbps, so we definitely have 3 zeros too much.

nE0sIghT:
It's megabytes per second in the documentation now 

image

https://github.com/Telmate/terraform-provider-proxmox/blob/master/docs/resources/vm_qemu.md


I have 768 in the hcl, 768 in the tfstate  and 768 in the Proxmox.

Tinyblargon:
@nE0sIghT Thank you for fixing the rate conversion

PR #1488, opened 2026-02-01
Change node to target_node in LXC guest resource

dsabella84:
Update to documentation instead of using node to be target_node for lxc_guest resource


Tinyblargon:
@dsabella84 Thank you for this contribution.

PR #1496, opened 2026-02-20
fix: #1339, state update / vm refresh slow

### LINKED ISSUE #1339: Takes long time to refresh VM states
Running provider `v3.0.2-rc01`

Symptom: When running `terraform plan` or `terraform apply`, the "Refreshing state" steps takes a long time.

From the debug log, I can see the `/network-get-interfaces` Proxmox API was being called repeatedly for some of the VMs. Those VMs are configured on a different subnet, not reachable from the host that runs `terraform plan`. The unreachable IP address was returned in the Proxmox API call via Qemu Agent.

Looks like the `getPrimaryIP()` function is checking the (unreachable) IP address and retrying several times before giving up.

How to bypass this IP address check? Thanks.

nutgood:
Fixes #1339, which surfaces a bug where `terraform plan` will block up to 90s per running VM.

`resourceVmQemuRead` polls the QEMU guest agent in a retry loop (up to 90s) to get the VM's IP. This makes sense after create/update when the agent is booting, but the same code runs during plan/refresh where there's no reason to wait. With N running VMs, a simple plan takes N * 90s in the worst case.

## Changes

Thread a `waitForAgent` bool through `resourceVmQemuRead -> initConnInfo -> getPrimaryIP`:

- Create/Update pass `true`, full retry loop as-is
- ReadContext passes `false` - agentTimeout is 0, single attempt no retry

### DIFF
--- a/proxmox/resource_vm_qemu.go
+++ b/proxmox/resource_vm_qemu.go
@@ -772,7 +772,7 @@ func resourceVmQemuCreate(ctx context.Context, d *schema.ResourceData, meta inte
 
 	d.Set("reboot_required", rebootRequired)
 	log.Print("[DEBUG][QemuVmCreate] vm creation done!")
-	return append(diags, resourceVmQemuRead(ctx, d, vmr, client)...)
+	return append(diags, resourceVmQemuRead(ctx, d, vmr, client, true)...)
 }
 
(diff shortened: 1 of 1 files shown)


Tinyblargon:
@nutgood tested it just now. Thank you for this contribution.

PR #1553, opened 2026-09-22
fix: don't let ReadVMHA failures abort qemu Create/Read

### LINKED ISSUE #1552: proxmox_vm_qemu: rc08's ReadVMHA call aborts Create()/Read() for any non-HA VM, silently, with zero diagnostic
### Terraform/Provider version

Terraform v1.11.6 (OpenTofu)

provider registry.terraform.io/telmate/proxmox v3.0.2-rc08

Proxmox VE 7.4-18 (also reproduced against a cluster on Proxmox VE 9.2.5).

### What happened

Any `proxmox_vm_qemu` created via `clone` (full clone, non-HA VM — i.e. almost
every VM) fails immediately after the clone itself completes successfully.
cores/memory/network/cloud-init are never applied. The only error surfaced is:

Error: api error: code: 500 message:

with an empty message, no further detail even with `TF_LOG=DEBUG`.

### Root cause

`v3.0.2-rc08` added an unconditional call to `client.ReadVMHA(ctx, vmr)` in
`resourceVmQemuRead` (`proxmox/resource_vm_qemu.go`, changelog entry "fix:
read hagroup value from Proxmox"). This issues `GET
/cluster/ha/resources/{vmid}`.

Proxmox answers this with a bare **HTTP 500** (not 404) for any VM that isn't
HA-managed — i.e. the normal case — per Proxmox's own bugzilla
[#5698](https://bugzilla.proxmox.com/show_bug.cgi?id=5698), which is still
open (confirmed against a Proxmox maintainer's own reply: the underlying
issue affects many endpoints, and is a version-independent quirk, not fixed
in any released PVE version). The response body is `{"data":null}` (13
bytes), with the real reason ("no such resource 'vm:<id>'") embedded only in
the HTTP status line's reason-phrase, which is not surfaced by the client.

This is the exact same underlying Proxmox behavior already tracked for LXC
in #1338 (there causing a hard panic: `interface conversion: interface {} is
nil, not *schema.Set`), but manifests differently here: on the QEMU side,
`resourceVmQemuRead`'s call site discards the error returned by
`client.ReadVMHA` entirely:

```go
// resource_vm_qemu.go, rc08
client.ReadVMHA(ctx, vmr)
and ReadVMHA itself (in Telmate/proxmox-api-go, proxmox/client.go)
does an unguarded type assertion after the request:


func (c *Client) ReadVMHA(ctx context.Context, vmr *VmRef) (err error) {
	var list map[string]interface{}
	url := fmt.Sprintf("/cluster/ha/resources/%d", vmr.vmId)
	err = c.GetJsonRetryable(ctx, url, &list, 3)
	if err == nil {
		list = list["data"].(map[string]interface{})   // panics if "data" is JSON null
		...
	}
	return
}
If GetJsonRetryable treats a 500 with a decodable JSON body as "no
transport error" (plausible, since {"data":null} decodes fine), err == nil holds despite the 500, and the following type assertion panics on the
nil data field — the same failure class as #1338, just resulting in a
clean SDK-recovered diagnostic here instead of a full crash trace. I
wasn't able to attach a debugger to confirm the exact panic vs. discarded-
error path with certainty (no live repro environment for this specific
issue), but the HTTP-level evidence (below) and the source read both point
here unambiguously.

Reproduction

  name        = "repro"
  target_node = "pve-node-a"
  clone       = "some-template"
  full_clone  = true
  bios        = "ovmf"

  cores  = 8
  memory = 32768

  disk {
    type    = "disk"
    slot    = "scsi0"
    storage = "local-zfs"
    size    = "32G"
  }

  network {
    id     = 0
    bridge = "vmbr0"
    model  = "virtio"
  }
  network {
    id     = 1
    bridge = "vmbr0"
    model  = "virtio"
  }
}
tofu apply output:


proxmox_vm_qemu.vm: Creating...
proxmox_vm_qemu.vm: Still creating... [10s elapsed]
proxmox_vm_qemu.vm: Still creating... [20s elapsed]
╷
│ Error: api error: code: 500 message: 
│ 
│   with proxmox_vm_qemu.vm,
│   on main.tf line 1, in resource "proxmox_vm_qemu" "vm":
│    1: resource "proxmox_vm_qemu" "vm" {
╵
pveproxy access.log for the same window (grep on the vmid), confirming
the exact failing call:


... "GET /api2/json/nodes/pve-node-a/qemu/123/config HTTP/1.1" 200 581
... "GET /api2/json/cluster/resources?type=vm HTTP/1.1" 200 123139
... "GET /api2/json/cluster/ha/resources/123 HTTP/1.1" 500 13
The resulting VM exists (the clone itself always succeeds) but keeps the
cloned template's defaults — cores/memory/network/cloud-init from the
resource config are never applied, since the operation aborts right after
this HA check and before the subsequent config-update step.

Fix
Downgrading to 3.0.2-rc07 (the last release before the ReadVMHA call was
added) resolves this completely — confirmed by rebuilding the same resource
against the same cluster/template with only the provider version changed.

Suggested fix
hagroup/hastate are optional, informational fields — a VM not being
HA-managed is the overwhelmingly common case and shouldn't be treated as a
fatal condition for Create()/Read(). I'm attaching a small patch that wraps
the ReadVMHA call so a failure (returned error or panic) is logged and
ignored rather than propagated. It compiles cleanly against v3.0.2-rc08
(go build ./..., gofmt -l clean) but I have not been able to verify it
end-to-end against a live cluster reproducing the panic path specifically —
happy to open this as a PR if a maintainer can confirm the approach, or if
someone with a repro handy can verify it fixes the panic case too (not just
the discarded-error case).

<details> <summary>Patch (proxmox/resource_vm_qemu.go)</summary>

@@ -937,7 +937,23 @@ func resourceVmQemuRead(ctx context.Context, d *schema.ResourceData, vmr *pveSDK
 	logger.Debug().Int(vmID.Root, int(vmr.VmId())).Msgf("[READ] Received Config from Proxmox API: %+v", config)
 
 	// NewActiveRawConfigQemuFromApi do not call ReadVMHA() so hagroup is not populated yet
-	client.ReadVMHA(ctx, vmr)
+	//
+	// Proxmox answers GET /cluster/ha/resources/{vmid} with a bare HTTP 500
+	// ("no such resource") instead of 404 whenever the VM isn't HA-managed -
+	// the normal case for most VMs (Proxmox bugzilla #5698, still open as of
+	// writing). Treat that as "no HA info available" instead of letting it
+	// abort the whole read, and guard against a possible panic decoding the
+	// resulting {"data":null} body.
+	func() {
+		defer func() {
+			if r := recover(); r != nil {
+				logger.Debug().Int(vmID.Root, int(vmr.VmId())).Msgf("[READ] ReadVMHA panicked, treating as not HA-managed: %v", r)
+			}
+		}()
+		if haErr := client.ReadVMHA(ctx, vmr); haErr != nil {
+			logger.Debug().Int(vmID.Root, int(vmr.VmId())).Msgf("[READ] ReadVMHA failed, treating as not HA-managed: %v", haErr)
+		}
+	}()
 
 	d.SetId(id.Guest{
 		ID:   vmr.VmId(),
</details>
References
Proxmox bugzilla [#5698](https://bugzilla.proxmox.com/show_bug.cgi?id=5698) — root API defect (500 instead of 404), confirmed open, version-independent.
#1338 — same underlying Proxmox behavior, LXC side, causes a hard panic there.
#486 — earlier, unrelated hagroup state-drift report (different symptom).

hackatoo:
Proxmox answers GET /cluster/ha/resources/{vmid} with a bare HTTP 500 ("no such resource") instead of 404 whenever the VM isn't HA-managed, which is the normal case for most VMs (Proxmox bugzilla #5698, still open, confirmed version-independent). Since 3.0.2-rc08 added an unconditional client.ReadVMHA(ctx, vmr) call in resourceVmQemuRead with its returned error discarded, this can abort the whole Create()/Read() before cores/memory/network/cloud-init are applied, surfacing only a bare "api error: code: 500 message: " with no further diagnostic.

hagroup/hastate are optional, informational fields, so treat a failure reading them (returned error or a possible panic decoding a {"data":null} body) as "not HA-managed" instead of fatal.

Closes #1552


Tinyblargon:
@hackatoo would you be okay with moving this change upstream to https://github.com/Telmate/proxmox-api-go ?

